PR #3, opened 2019-10-03
Feature/checkout

zunkas:


src/SwedbankPay.Sdk/Models/Callbacks/Callback.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace SwedbankPay.Client.Models.Callbacks
+﻿namespace SwedbankPay.Sdk.Models.Callbacks
```
asbjornu: I don't think `Models` provide any useful information in a namespace. Everything that is a `class`, `struct` or `interface` is a model. Please remove it.

src/SwedbankPay.Sdk/Models/Common/AccountInfo.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace SwedbankPay.Client.Models.Common
+﻿namespace SwedbankPay.Sdk.Models.Common
```
asbjornu: I don't think `Common` provide any useful information in a namespace. Everything "parent" to the namespace scope you're in is "common" to the current scope. Just removing `Models.Common` will make everything within the `Common` namespace more common by simply being more accessible since it will be implicitly included by all namespaces starting with `SwedbankPay.Sdk`. Please remove it.

src/SwedbankPay.Sdk/Models/Request/ConsumersRequest.cs
```diff
@@ -1,10 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
-namespace SwedbankPay.Client.Models.Request
+﻿namespace SwedbankPay.Sdk.Models.Request
```
asbjornu: Please don't group by type, but by feature. Everything related to consumer should be within a `Consumers` namespace. Having all `Request` classes grouped together is nonsensical. Please change it so that regardless of type, everything related to consumer is within a `Consumers` namespace, everything related to credit card is within a `CreditCard` namespace, and so on.

src/SwedbankPay.Sdk/Models/Response/ConsumersResponse.cs
```diff
@@ -1,11 +1,7 @@
...
-using System.Threading.Tasks;
-
-namespace SwedbankPay.Client.Models.Response
+﻿namespace SwedbankPay.Sdk.Models.Response
 {
+    using System.Collections.Generic;
+
     public class ConsumersResponse
     {
         public string Token { get; set; }
```
asbjornu: Why does a `Response` object have a `public` setter for any of its properties? What does it mean to a consumer of `ConsumerResponse` that `ConsumerResponse.Token` can be changed? Please ensure immutability (preferably with `readonly`, but at least with `private` setters) wherever it's possible to enforce.

src/SwedbankPay.Sdk/Models/Vipps/TransactionAPI/Response/AllTransactionResponseContainer.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace PayEx.Client.Models.Vipps.TransactionAPI.Response
+﻿namespace SwedbankPay.Sdk.Models.Vipps.TransactionAPI.Response
```
asbjornu: Why the name `TransactionAPI`? I would like this namespace to read:

```suggestion
namespace SwedbankPay.Sdk.Vipps.Transactions
```

Or perhaps without `Transactions` altogether and just `SwedbankPay.Sdk.Vipps`.

src/SwedbankPay.Sdk/Resources/ConsumerResource.cs
```diff
@@ -0,0 +1,60 @@
+﻿namespace SwedbankPay.Sdk.Resources
```
asbjornu: Please delete the `Resources` namespace and move all classes within it alongside the feature they relate to.

src/SwedbankPay.Sdk/Resources/ConsumerResource.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    using System;
+    using System.Threading.Tasks;
+    using RestSharp;
+    using SwedbankPay.Sdk.Exceptions;
+    using SwedbankPay.Sdk.Models;
+    using SwedbankPay.Sdk.Models.Request;
+    using SwedbankPay.Sdk.Models.Response;
+
+    public class ConsumerResource : ResourceBase, IConsumerResource
```
asbjornu: `ConsumerResource` belongs inside the `Consumers` namespace alongside everything else related to consumers.

src/SwedbankPay.Sdk/Resources/ConsumerResource.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            var url = "/psp/consumers";
+            
+            Func<ProblemsContainer, Exception> onError = m => new CouldNotInitiateConsumerSessionException(consumersRequest, m);
+            consumersRequest.Operation = "initiate-consumer-session";
+            var res = await CreateInternalClient().HttpPost<ConsumersRequest, ConsumersResponse>(url, onError, consumersRequest);
+           
+            return res;
+        }
+        public async Task<ShippingDetails> GetShippingDetails(string uri)
```
asbjornu: Where does the URI for shipping details originate from and why isn't it typed as a `System.Uri`?

src/SwedbankPay.Sdk/Resources/ConsumerResource.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+            Func<ProblemsContainer, Exception> onError = m => new CouldNotGetShippingDetailsException(uri, m);
+            var request = new RestRequest(uri, Method.GET);
+            
+            var shippingDetails = await CreateInternalClient().HttpGet<ShippingDetails>(uri, onError);
+
+            return shippingDetails;
+        }
+
+        public async Task<BillingDetails> GetBillingDetails(string uri)
```
asbjornu: Where does the URI for billing details originate from and why isn't it typed as a `System.Uri`?

src/SwedbankPay.Sdk/Resources/IPaymentOrdersResource.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    using System.Threading.Tasks;
+    using SwedbankPay.Sdk.Exceptions;
+    using SwedbankPay.Sdk.Models;
+    using SwedbankPay.Sdk.Models.Request;
+    using SwedbankPay.Sdk.Models.Request.Transaction;
+    using SwedbankPay.Sdk.Models.Response.PaymentOrder;
+    using SwedbankPay.Sdk.Models.Vipps.TransactionAPI.Response;
+
+    public interface IPaymentOrdersResource
```
asbjornu: This interface belongs in the `PaymentOrders` namespace.

src/SwedbankPay.Sdk/Resources/IPaymentResource.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using SwedbankPay.Sdk.Models;
+    using SwedbankPay.Sdk.Models.Request.Transaction;
+    using SwedbankPay.Sdk.Models.Response.Payment;
+    using SwedbankPay.Sdk.Models.Vipps.PaymentAPI.Request;
+    using SwedbankPay.Sdk.Models.Vipps.TransactionAPI.Response;
+
+    public interface IPaymentResource
```
asbjornu: This interface belongs in the `Payments` namespace.

src/SwedbankPay.Sdk/ILogPayExHttpResponse.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace PayEx.Client
+﻿namespace SwedbankPay.Sdk
 {
     public interface ILogPayExHttpResponse
```
asbjornu:
```suggestion
    public interface ILogSwedbankPayHttpResponse
```

src/SwedbankPay.Sdk/Utils.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace SwedbankPay.Sdk
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Serialization;
+    using RestSharp;
+
+    internal static class Utils
```
asbjornu: Please delete this class. We should not need to have any `Utils` in the code base. By using proper object oriented programming, all methods within this class can be moved into base classes and specific classes where they belong.

src/SwedbankPay.Sdk/Utils.cs
```diff
@@ -0,0 +1,57 @@
...
+                {
+                    s.Add(name.ToLower());
+                }
+            }
+
+            var queryString = string.Join(",", s);
+            return $"?$expand={queryString}";
+        }
+
+        internal static THeaderValue GetHeaderResponseValue<THeaderValue>(this IRestResponse response, string name) where THeaderValue : class
```
asbjornu: Where is `IRestResponse` defined?

zunkas: In RestSharp, this method is neved used and will be removed since Utils.cs should be removed

src/SwedbankPay.Sdk/Resources/ConsumerResource.cs
zunkas: It's from ConsumersResource "On Shipping Details Available" 
{
    "actionType": "OnBillingDetailsAvailable",
    "url": "/psp/consumers/<consumerProfileRef>/billing-details"
}

Renaming to url instead.

src/SwedbankPay.Sdk/Resources/ConsumerResource.cs
zunkas: It's from ConsumersResource "On Billing Details Available" 
{
    "actionType": "OnBillingDetailsAvailable",
    "url": "/psp/consumers/<consumerProfileRef>/billing-details"
}

Renaming to url instead.

src/SwedbankPay.Sdk/Payments/Price.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace SwedbankPay.Sdk.Payments
+{
+    public class Price
+    {
+        public string Type { get; set; }
```
asbjornu: Is it possible to represent `Type` as an `enum`?

src/SwedbankPay.Sdk/Payments/SaleResponse.cs
```diff
@@ -0,0 +1,18 @@
...
+    using System;
+
+    public class SaleResponse
+    {
+        public DateTime Date { get; protected set; }
+        public string PayerAlias { get; protected set; }
+        public string SwishPaymentReference { get; protected set; }
+        public string SwishStatus { get; protected set; }
+
+        public string Id { get; protected set; }
```
asbjornu: Isn't `Id` a `System.Uri`?

src/SwedbankPay.Sdk/ResourceBase.cs
```diff
@@ -0,0 +1,71 @@
...
+    using System.Threading.Tasks;
+
+    public abstract class ResourceBase
+    {
+        protected readonly SwedbankPayOptions SwedbankPayOptions;
+        private readonly ILogSwedbankPayHttpResponse _logger;
+
+        internal ResourceBase(SwedbankPayOptions swedbankPayOptions, ILogSwedbankPayHttpResponse logger)
+        {
+            SwedbankPayOptions = swedbankPayOptions;
```
asbjornu: What happens if `swedbankPayOptions` is `null`? All constructors need proper argument guards.

```suggestion
            SwedbankPayOptions = swedbankPayOptions ?? throw new ArgumentNullException(nameof(swedbankPayOptions));
```

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,136 @@
+﻿namespace SwedbankPay.Sdk
+{
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+    using Newtonsoft.Json.Serialization;
+
+    using RestSharp;
```
asbjornu: Please don't use `RestSharp`. Microsoft's `HttpClient` provides a much better abstraction for what this SDK is supposed to do.

src/SwedbankPay.Sdk/Transactions/TransactionListContainer.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace SwedbankPay.Sdk.Transactions
+{
+    using System.Collections.Generic;
+
+    public class TransactionListContainer : IdLink
+    {
+        public List<TransactionResponse> TransactionList { get; protected set; } = new List<TransactionResponse>();
+    }
+
+    public class AuthorizationListContainer : IdLink
```
asbjornu: Only one class per file, please.

src/SwedbankPay.Sdk/Transactions/ReversalTransactionResponseContainer.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SwedbankPay.Sdk.Transactions
+{
+    internal class ReversalTransactionResponseContainer
+    {
+        public string Payment { get; protected set; }
```
asbjornu: Isn't `Payment` a `System.Uri`?

src/SwedbankPay.Sdk/Transactions/CaptureTransactionResponseContainer.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SwedbankPay.Sdk.Transactions
+{
+    internal class CaptureTransactionResponseContainer
+    {
+        public string Payment { get; protected set; }
```
asbjornu: Isn't `Payment` a `System.Uri`?

src/SwedbankPay.Sdk/Transactions/CancellationTransactionResponseContainer.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SwedbankPay.Sdk.Transactions
+{
+    internal class CancellationTransactionResponseContainer
+    {
+        public string Payment { get; protected set; }
```
asbjornu: Isn't `Payment` a `System.Uri`?

src/SwedbankPay.Sdk/Transactions/AllTransactionResponseContainer.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace SwedbankPay.Sdk.Transactions
+{
+    internal class AllTransactionResponseContainer
+    {
+        public string Payment { get; protected set; }
```
asbjornu: Isn't `Payment` a `System.Uri`?

src/SwedbankPay.Sdk/SwedbankPayClient.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    public class SwedbankPayClient : ISwedbankPayClient
+    {
+        public IPaymentOrdersResource PaymentOrders { get; }
+        public IPaymentsResource Payment { get; }
+        public IConsumersResource Consumers { get; }
+
+        public SwedbankPayClient(SwedbankPayOptions swedbankPayOptions, ILogSwedbankPayHttpResponse logger = null)
+        {
+            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
```
asbjornu: Is it possible to set `SecurityProtocol` to `Tls12` only if it is set to something lower than `Tls12` and not otherwise?

src/SwedbankPay.Sdk/PaymentOrders/PayeeInfo.cs
```diff
@@ -16,6 +16,22 @@ public class PayeeInfo
         /// </summary>
         public string PayeeReference { get; set; }
 
-        public string PayeeName { get; set; } 
+        public string PayeeName { get; set; }
```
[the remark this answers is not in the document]
karlfristedt76: image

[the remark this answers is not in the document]
asbjornu: @karlfristedt76, you can just perform the changes in code, no need to tell me about it. :smiley: :+1:

src/SwedbankPay.Sdk/Consumers/ConsumersRequest.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    using SwedbankPay.Sdk.PaymentOrders;
+
+    public class ConsumersRequest
+    {
+        public string Operation { get; internal set; }
+        public string Msisdn { get; set; }
+        public string Email { get; set; }
+
+        public string ConsumerCountryCode { get; set; }
```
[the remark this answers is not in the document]
karlfristedt76: That sounds like a good idea. We just need info on valid countrycodes. NO and SE are obvious but are there more?

src/Samples/Sample.AspNetCore/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+    public static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection AddSwedbankPayClient(this IServiceCollection services, IConfiguration configuration, string clientName)
+        {
+            var swedbankPayConfigsection = configuration.GetSection($"SwedbankPay:{clientName}");
+            services.Configure<SwedbankPayOptions>(clientName, swedbankPayConfigsection);
+            var swedbankPayOptions = swedbankPayConfigsection.Get<SwedbankPayOptions>();
+            services.AddTransient<SwedbankPayOptions>();
+            services.AddTransient(s => new SwedbankPayClient(swedbankPayOptions));
```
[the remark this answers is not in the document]
karlfristedt76: Its not necessary in this case. This is an old sample from ICE perhaps we should remove this project and instead use the new Aspnetcore sample site?

src/Samples/Sample.ConsoleApp/Program.cs
```diff
@@ -16,81 +15,29 @@ static void Main(string[] args)
...
-        public string Select()
-        {
-            return Constants.THECLIENTNAME;
-        }
-    }
 
-    internal class HttpClientCreator : IHttpClientFactory
-    {
-        private readonly PayExOptions _payExOptions;
+            var res = swedbankPayClient.Payment.PostVippsPayment(paymentRequest).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
karlfristedt76: This is also an old sample project from ICE

src/Samples/Sample.ConsoleApp/Program.cs
```diff
@@ -16,81 +15,29 @@ static void Main(string[] args)
...
-        {
-            _payExOptions = payExOptions;
-        }
-        public HttpClient CreateClient(string name)
-        {
-            var httpClient = new HttpClient();
-            httpClient.DefaultRequestHeaders.Add("Authorization", $"Bearer {_payExOptions.Token}");
-            httpClient.BaseAddress = _payExOptions.ApiBaseUrl;
-            return httpClient;
+            var dynamicallyCreated = swedbankPayClient.Payment.PostVippsPayment(paymentRequest).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
karlfristedt76: This is also an old sample project from ICE

src/SwedbankPay.Sdk/PaymentOrders/CurrentPaymentResponseContainer.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace SwedbankPay.Sdk.PaymentOrders
+{
+    using SwedbankPay.Sdk.Payments;
+
+    public class CurrentPaymentResponseContainer : PaymentResponseContainer
+    {
+        public string Id { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/SwedbankPay.Sdk/PaymentOrders/Address.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace SwedbankPay.Sdk.PaymentOrders
+{
+    public class Address
+    {
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+        public string Email { get; set; }
+        public string Msisdn { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/SwedbankPay.Sdk/PaymentOrders/Address.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace SwedbankPay.Sdk.PaymentOrders
+{
+    public class Address
+    {
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+        public string Email { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/SwedbankPay.Sdk/PaymentOrders/AccountInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// 02 (Account name different than shipping name)
+        /// </summary>
+        public string ShippingNameIndicator { get; set; }
+
+        /// <summary>
+        /// Indicates whether merchant has experienced suspicious activity (including previous fraud) on the cardholder account.
+        /// 01 (No suspicious activity has been observed)
+        /// 02 (Suspicious activity has been observed)
+        /// </summary>
+        public string SuspiciousAccountActivity { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/SwedbankPay.Sdk/PaymentOrders/AccountInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// 04 (More than 60 days)
+        /// </summary>
+        public string ShippingAddressUsageIndicator { get; set; }
+
+        /// <summary>
+        /// Indicates if the Cardholder Name on the account is identical to the shipping Name used for this transaction.
+        /// 01 (Account name identical to shipping name)
+        /// 02 (Account name different than shipping name)
+        /// </summary>
+        public string ShippingNameIndicator { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/SwedbankPay.Sdk/PaymentOrders/AccountInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+        public string AccountPwdChangeIndicator { get; set; }
+
+        /// <summary>
+        /// Indicates when the shipping address used for this transaction was first used with the merchant.
+        /// 01 (This transaction)
+        /// 02 (Less than 30 days)
+        /// 03 (30-60 days)
+        /// 04 (More than 60 days)
+        /// </summary>
+        public string ShippingAddressUsageIndicator { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/SwedbankPay.Sdk/PaymentOrders/AccountInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Indicates the length of time since the cardholder's account with the merchant had a password change or account reset.
+        /// 01 (No change)
+        /// 02 (Changed during transaction)
+        /// 03 (Less than 30 days)
+        /// 04 (30-60 days)
+        /// 05 (More than 60 days)
+        /// </summary>
+        public string AccountPwdChangeIndicator { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/SwedbankPay.Sdk/PaymentOrders/AccountInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+        public string AccountAgeIndicator { get; set; }
+
+        /// <summary>
+        /// Length of time since the cardholder's account information with the merchant was changed. Including billing etc.
+        /// 01 (Changed during transaction)
+        /// 02 (Less than 30 days)
+        /// 03 (30-60 days)
+        /// 04 (More than 60 days)
+        /// </summary>
+        public string AccountChangeIndicator { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/SwedbankPay.Sdk/PaymentOrders/AccountInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        /// <summary>
+        /// Indicates the length of time that the payments account was enrolled in the cardholder's account with merchant.
+        /// 01 (No account, guest)
+        /// 02 (Created during transaction)
+        /// 03 (Less than 30 days)
+        /// 04 (30-60 days)
+        /// 05 (More than 60 days)
+        /// </summary>
+        public string AccountAgeIndicator { get; set; }
```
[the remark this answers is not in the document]
asbjornu: @karlfristedt76 ☝️

src/Samples/Sample.AspNetCore/ServiceCollectionExtensions.cs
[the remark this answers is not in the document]
asbjornu: @karlfristedt76, what does the old ICE sample provide that our new AspNetCore site doesn't? If our new sample is more feature-rich, I see no reason to keep the old one.

src/SwedbankPay.Sdk/Payments/IPaymentsResource.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// Gets all transactions for a payment.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
+        Task<IEnumerable<TransactionResponse>> GetTransactions(string id, PaymentExpand paymentExpand = PaymentExpand.All);
+
+        /// <summary>
+        /// Captures a payment a.k.a POSTs a transaction.
+        /// </summary>
+        Task<TransactionResponse> PostCapture(string id, TransactionRequest transaction);
```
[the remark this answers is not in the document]
karlfristedt76: Lets discuss this next meeting.

src/SwedbankPay.Sdk/ResourceBase.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using SwedbankPay.Sdk.Exceptions;
+using SwedbankPay.Sdk.PaymentOrders;
+
+namespace SwedbankPay.Sdk
+{
+    public abstract class ResourceBase
+    {
+        protected readonly SwedbankPayOptions swedbankPayOptions;
+        internal SwedbankPayHttpClient swedbankPayClient;
```
muzukashiine: There already exists a class with the name SwedbankPayClient which makes this confusing. Should be renamed to swedbankPayHttpClient.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrder.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+
+        internal static async Task<PaymentOrder> Get(string id,
+                                                     SwedbankPayHttpClient client,
+                                                     SwedbankPayOptions swedbankPayOptions,
+                                                     PaymentOrderExpand paymentOrderExpand = PaymentOrderExpand.None)
+        {
+            if (string.IsNullOrEmpty(id))
+                throw new CouldNotFindPaymentException(id);
```
muzukashiine: Isn't this quite misleading? It's an input check error, the payment order might still exist?

src/SwedbankPay.Sdk/ResourceBase.cs
```diff
@@ -0,0 +1,69 @@
...
+                throw new ArgumentNullException(nameof(swedbankPayOptions));
+
+            if (client.BaseAddress == null || !client.DefaultRequestHeaders.Contains("Authorization"))
+                throw new ArgumentException("Invalid client configuration. Check config.");
+
+            if (swedbankPayOptions.IsEmpty())
+                throw new InvalidConfigurationSettingsException("Invalid configuration. Check config.");
+
+            this.swedbankPayOptions = swedbankPayOptions;
+            this.swedbankPayClient = new SwedbankPayHttpClient(client, logger);
```
muzukashiine: Probably a better idea to pass this in instead of instantiating it here?

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,227 @@
...
+                problems = new ProblemsContainer("id", "Not found");
+            else
+                problems = new ProblemsContainer("Other", $"Response when calling SwedbankPay was: '{response.StatusCode}'");
+
+            var ex = onError(problems);
+            throw ex;
+        }
+
+
+        internal async Task<T> HttpRequest
```
muzukashiine: 90% of this method is the same as the other HttpRequest() method.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,227 @@
...
+                return JsonConvert.DeserializeObject<T>(res, JsonSerialization.JsonSerialization.Settings);
+            }
+
+            var responseMessage = await response.Content.ReadAsStringAsync();
+            this.logger.LogInformation(responseMessage);
+            ProblemsContainer problems;
+            if (!string.IsNullOrEmpty(responseMessage) && IsValidJson(responseMessage))
+                problems = JsonConvert.DeserializeObject<ProblemsContainer>(responseMessage);
+            else if (response.StatusCode == HttpStatusCode.NotFound)
+                problems = new ProblemsContainer("id", "Not found");
```
muzukashiine: How far do we want to abstract this? Why not just let the HTTP exception bubble to the top, including the common 404 Not Found?

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,227 @@
...
+            where T : new()
+        {
+            UpdateRequest(requestMessage, payload);
+
+            HttpResponseMessage response;
+            try
+            {
+                response = await this.client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead).ConfigureAwait(false);
+            }
+            catch (HttpRequestException e)
```
muzukashiine: I feel these catch-clauses are obscuring the exception flow with unnecessary fluff. What's the difference between letting a TaskCanceledException bubble to the top instead of throwing it as an ApiTimeOutException? When the JSON parser is deserializing the content we don't catch and re-throw that, for example. I think the caller will have a better chance at understanding what went wrong if we allow the original exceptions to get thrown instead.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,227 @@
...
+        internal async Task<T> HttpRequest
+            <T>(string httpMethod, string url, Func<ProblemsContainer, Exception> onError, object payload = null)
+            where T : new()
+        {
+            return await HttpRequest<T>(new HttpMethod(httpMethod), url, onError, payload);
+        }
+
+
+        internal async Task<T> HttpRequest
+            <T>(HttpMethod httpMethod, string url, Func<ProblemsContainer, Exception> onError, object payload = null)
```
muzukashiine: What advantages does the onError() approach give us? Why not just let the caller handle this themselves by catching whatever we throw at them?

src/SwedbankPay.Sdk.Tests/TestHelpers/TestHelper.cs
```diff
@@ -0,0 +1,86 @@
...
+            return new ConfigurationBuilder()
+                .SetBasePath(outputPath)
+                .AddJsonFile("appsettings.json", true)
+                .AddUserSecrets("55739ea0-5447-45e4-b35e-e0412f172f5f")
+                .AddEnvironmentVariables()
+                .AddInMemoryCollection(new Dictionary<string, string>
+                {
+                    { "SwedbankPayConnectionSettings:Token", Environment.GetEnvironmentVariable("SwedbankPayConnectionSettings:Token")},
+                    { "SwedbankPayConnectionSettings:ApiBaseUrl", Environment.GetEnvironmentVariable("SwedbankPayConnectionSettings:ApiBaseUrl") },
+                    { "Urls:TermsOfServiceUrl", Environment.GetEnvironmentVariable("Urls:TermsOfServiceUrl") },
```
muzukashiine: What's the purpose of adding this code twice?

src/SwedbankPay.Sdk.Tests/TestHelpers/TestHelper.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.Extensions.Configuration;
+
+namespace SwedbankPay.Sdk.Tests.TestHelpers
+{
+    public class TestHelper
+    {
+
+        public static IConfigurationRoot GetIConfigurationRoot(string outputPath)
+        {
+#if DEBUG
```
muzukashiine: Consider removing this if-else as it can lead to unexpected behavior when using the SDK package. Just imagine what would happen if a project using the SDK is run locally in debug mode, for example.

src/SwedbankPay.Sdk.Tests/TestHelpers/TestHelper.cs
```diff
@@ -0,0 +1,86 @@
...
+            return new ConfigurationBuilder()
+                .SetBasePath(outputPath)
+                .AddJsonFile("appsettings.json", true)
```
muzukashiine: Consider using either appsettings or environment variables, but not both.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,193 @@
...
+        internal async Task<TResponse> HttpPost
+            <TPayLoad, TResponse>(string url, Func<ProblemsContainer, Exception> onError, TPayLoad payload)
+            where TResponse : new()
+        {
+            return await HttpRequest<TResponse>(HttpMethod.Post, url, onError, payload);
+        }
+
+
+        internal async Task<T> HttpRequest
+            <T>(string httpMethod, string url, Func<ProblemsContainer, Exception> onError, object payload = null)
```
muzukashiine: The only difference from the other HttpRequest() is that this takes a httpMethod input as a string. I can't see why this would be useful enough to warrant its own method when the caller could just wrap the string in a HttpMethod object themselves.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,193 @@
...
+            where T : new()
+        {
+            return await HttpRequest<T>(new HttpMethod(httpMethod), url, onError, payload);
+        }
+
+
+        internal async Task<T> HttpRequest
```
muzukashiine: "HttpRequest" is not a very descriptive method name. How about "SendHttpRequestAndProcessHttpResponse" or something along those lines?

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrder.cs
```diff
@@ -0,0 +1,122 @@
...
+                return new CouldNotFindPaymentException(id, m);
+            }
+
+            var paymentOrderResponseContainer = await client.HttpGet<PaymentOrderResponseContainer>(url, OnError);
+
+            return new PaymentOrder(paymentOrderResponseContainer, client);
+        }
+
+
+        internal static string GetExpandQueryString<T>(T expandParameter)
```
muzukashiine: PaymentOrderExpand is the only enum being used here so consider using that as the input type instead of generic T.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderExpand.cs
```diff
@@ -0,0 +1,19 @@
...
+        None = 0,
+        Urls = 1,
+        PayeeInfo = 2,
+        Settings = 4,
+        Payers = 8,
+        OrderItems = 16,
+        MetaData = 32,
+        Payments = 64,
+        CurrentPayment = 128,
+        All = 255
```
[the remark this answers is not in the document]
muzukashiine: I agree. Could be something as simple as a `PaymentOrderExpand` class that one can call methods on to gradually build up a list with predefined values, and then call `ToString()` which joins the list into a string. Currently the handling of this bitfield-enum (in PaymentOrder.cs) is messy and confusing and I'd advise against it.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrder.cs
muzukashiine: Or even better, don't use an enum with `[Flags]` and instead go for the solution suggested by @asbjornu above.

src/SwedbankPay.Sdk/LinkRelation.cs
```diff
@@ -0,0 +1,65 @@
...
+        public static readonly LinkRelation CreateReversal =
+            new LinkRelation(nameof(CreateReversal), PaymentResourceOperations.CreateReversal);
+
+        public static readonly LinkRelation PaidPaymentOrder =
+            new LinkRelation(nameof(PaidPaymentOrder), PaymentOrderResourceOperations.PaidPaymentOrder);
+
+        public static readonly LinkRelation PaidPayment = new LinkRelation(nameof(PaidPayment), PaymentResourceOperations.PaidPayment);
+
+
+        public LinkRelation(string name, string value)
```
muzukashiine: public -> private

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrder.cs
```diff
@@ -0,0 +1,122 @@
...
+                    case PaymentOrderResourceOperations.UpdatePaymentOrderUpdateOrder: 
+                        operations.Update = new ExecuteRequestWrapper<PaymentOrderUpdateRequestContainer, PaymentOrderResponseContainer>(
+                            httpOperation.Request, client, m => new CouldNotUpdatePaymentOrderException(httpOperation.Href, m));
+                        break;
+                    case PaymentOrderResourceOperations.UpdatePaymentOrderAbort:
+                        operations.Abort = new ExecuteWrapper<PaymentOrderResponseContainer>(
+                            httpOperation.Request, client, m => new CouldNotPostTransactionException(httpOperation.Href, m),
+                            new PaymentAbortRequestContainer());
+                        break;
+                    case "view-paymentorder":
```
muzukashiine: Use static string/enum

src/SwedbankPay.Sdk.Tests/TestHelpers/TestHelper.cs
zunkas: This is the Test project, this have nothing to do with how the SDK is run.

src/SwedbankPay.Sdk.Tests/TestHelpers/TestHelper.cs
zunkas: This is for solving how devops fetches the e.g. token so we don't need to have it in the appsettings.json file in git.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrder.cs
zunkas: Changed to use the GetExpandQueryString from ResourceBase instead of the method in PaymentOrder. So it's still generic but uses more than PaymentOrderExpand.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
zunkas: OnError throws the exception that the SDK tells it to throw. For instance when a request is successful but the api returns a "Problem". E.g.` PaymentOrder.Create` throws a `CouldNotPlacePaymentOrderException`

It's still up to the caller to handle this exception themselves.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
zunkas: Here it's a successful response with no Exception? Just a "ProblemResponse"? 
And that's why OnError throws a exception instead.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderExpand.cs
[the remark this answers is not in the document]
muzukashiine: Something like this:

`public abstract class ExpandBuilder
    {
        private readonly List<string> list;

        protected ExpandBuilder()
        {
            list = new List<string>();
        }

        public void AddMetadata()
        {
            // add "metadata" to list
        }

        public void AddSettings()
        {
            // add "settings" to list
        }

        public override string ToString()
        {
            // remove duplicates
            // join list into string
        }
    }`

src/SwedbankPay.Sdk/LinkRelation.cs
```diff
@@ -0,0 +1,65 @@
...
+// 
+// This software is proprietary information of Swedbank Pay.
+// USE IS SUBJECT TO LICENSE TERMS.
+// --------------------------------------------------
+
+#endregion
+
+namespace SwedbankPay.Sdk
+{
+    public sealed class LinkRelation : TypeSafeEnum<LinkRelation, string>
```
muzukashiine: Why do we need this abstraction layer?

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        internal async Task<TResponse> HttpPost
+            <TPayLoad, TResponse>(string url, Func<ProblemsContainer, Exception> onError, TPayLoad payload)
+            where TResponse : new()
+        {
+            return await SendHttpRequestAndProcessHttpResponse<TResponse>(HttpMethod.Post, url, onError, payload);
+        }
+
+
+        internal async Task<T> SendHttpRequestAndProcessHttpResponse
```
muzukashiine: What is "T"? Consider renaming it to something more descriptive such as TResponse.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+            <T>(HttpRequestMessage requestMessage, Func<ProblemsContainer, Exception> onError, object payload = null)
+            where T : new()
+        {
+            UpdateRequest(requestMessage, payload);
+
+            HttpResponseMessage response = await this.client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead).ConfigureAwait(false);
```
muzukashiine: HttpResponseMessage -> var

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+            
+            if (response.IsSuccessStatusCode)
+            {
+                var res = await response.Content.ReadAsStringAsync();
+                this.logger.LogInformation(res);
+                return JsonConvert.DeserializeObject<T>(res, JsonSerialization.JsonSerialization.Settings);
+            }
+
+            var responseMessage = await response.Content.ReadAsStringAsync();
```
muzukashiine: These two lines are duplicates of the ones inside the IF-clause above.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+            this.logger.LogInformation(responseMessage);
+            ProblemsContainer problems;
+            if (!string.IsNullOrEmpty(responseMessage) && IsValidJson(responseMessage))
+                problems = JsonConvert.DeserializeObject<ProblemsContainer>(responseMessage);
+            else if (response.StatusCode == HttpStatusCode.NotFound)
+                problems = new ProblemsContainer("id", "Not found");
+            else
+                problems = new ProblemsContainer("Other", $"Response when calling SwedbankPay was: '{response.StatusCode}'");
+
+            var ex = onError(problems);
```
muzukashiine: HTTP response should be part of the thrown exception.

src/SwedbankPay.Sdk/LinkRelation.cs
zunkas: Because LinkRelation is used as an Enum, and Operations inherits from Dictionary<LinkRelation, HttpOpertation> 
This is something we changed in discussion with @asbjornu

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        internal async Task<T> SendHttpRequestAndProcessHttpResponse
+            <T>(HttpRequestMessage requestMessage, Func<ProblemsContainer, Exception> onError, object payload = null)
+            where T : new()
+        {
+            UpdateRequest(requestMessage, payload);
+
+            HttpResponseMessage response = await this.client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead).ConfigureAwait(false);
+            
+            if (response.IsSuccessStatusCode)
```
muzukashiine: Consider adding a try-clause so that if something explodes in here we can still catch it and attach the HTTP response. Something as innocent as parsing JSON may fail.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
muzukashiine: I don't think we should return a ProblemsContainer for anything that isn't an actual ProblemJson. By checking for Not Found and creating a ProblemsContainer like we're doing here we're throwing away valuable information that may reside in the HTTP response content, for example.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                var res = await response.Content.ReadAsStringAsync();
+                this.logger.LogInformation(res);
+                return JsonConvert.DeserializeObject<T>(res, JsonSerialization.JsonSerialization.Settings);
+            }
+
+            var responseMessage = await response.Content.ReadAsStringAsync();
+            this.logger.LogInformation(responseMessage);
+            ProblemsContainer problems;
+            if (!string.IsNullOrEmpty(responseMessage) && IsValidJson(responseMessage))
```
muzukashiine: IsNullOrEmpty -> IsNullOrWhitespace

muzukashiine: What if we receive an erronous HTTP status code but with valid JSON attached? There's no guarantee the contents will be deserializable to the ProblemsContainer type and this will explode.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+        internal async Task<TResponse> HttpPatch
+            <TPayLoad, TResponse>(string url, Func<ProblemsContainer, Exception> onError, TPayLoad payload)
+            where TResponse : new()
+        {
+            return await SendHttpRequestAndProcessHttpResponse<TResponse>(new HttpMethod("PATCH"), url, onError, payload);
+        }
+
+
+        internal async Task<TResponse> HttpPost
+            <TPayLoad, TResponse>(string url, Func<ProblemsContainer, Exception> onError, TPayLoad payload)
```
muzukashiine: Here it is `TPayload` but below it is `object payload`.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+
+
+        internal SwedbankPayHttpClient(HttpClient client, ILogger logger)
+        {
+            this.client = client;
+            this.logger = logger;
+        }
+
+
+        internal async Task<string> GetRaw(string url)
```
muzukashiine: All comments below also apply to this method.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+                problems = new ProblemsContainer("id", "Not found");
+            else
+                problems = new ProblemsContainer("Other", $"Response when calling SwedbankPay was: '{response.StatusCode}'");
+
+            var ex = onError(problems);
+            throw ex;
+        }
+
+
+        private string GetRequestBody(object request)
```
muzukashiine: Not used anywhere. Remove?

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+            where TResponse : new()
+        {
+            var requestMessage = new HttpRequestMessage(httpMethod, url);
+
+            return await SendHttpRequestAndProcessHttpResponse<TResponse>(requestMessage, onError, payload);
+        }
+
+
+        internal async Task<TResponse> SendHttpRequestAndProcessHttpResponse
+            <TResponse>(HttpRequestMessage requestMessage, Func<ProblemsContainer, Exception> onError, object payload = null)
```
muzukashiine: Consider taking in only the requestMessage and not the payload as well. It's really messy having to send around two arguments that are only going to get merged further down the road anyway.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,173 @@
...
+                var res2 = JToken.Parse(res).ToString(Formatting.Indented);
+                this.logger.LogInformation(res2);
+                return res2;
+            }
+
+            return null;
+        }
+
+
+        internal async Task<TResponse> HttpGet<TResponse>(string url, Func<ProblemsContainer, Exception> onError)
```
muzukashiine: The URL should be a Uri object.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
muzukashiine: That still doesn't explain why using the "OnError approach" is advantageous to simply throwing. As a side note OnError should be renamed to something like ExceptionToBeThrownOnError, perhaps? Is the lambda meant to actually do something in addition to returning an exception, or just return an exception?

src/SwedbankPay.Sdk/Consumers/ConsumersResource.cs
```diff
@@ -0,0 +1,62 @@
...
+        ///     Retrieve Consumer Billing Details.
+        ///     When the payer has been identified through checkin you can retrieve the consumers billing details with the url
+        ///     received through the event OnBillingDetailsAvailable.
+        /// </summary>
+        /// <param name="url"></param>
+        /// <returns></returns>
+        public async Task<BillingDetails> GetBillingDetails(Uri url)
+        {
+            if (url == null)
+                throw new ArgumentNullException(nameof(url), "url cannot be null or whitespace");
```
muzukashiine: The check says nothing about whitespace so I'd suggest removing that comment. Also the exception type is quite self-descriptive so in my opinion there's no need for a comment at all.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,170 @@
...
+        internal async Task<TResponse> SendHttpRequestAndProcessHttpResponse
+            <TResponse>(HttpRequestMessage httpRequest)
+            where TResponse : new()
+        {
+            var httpResponse = await this.client.SendAsync(httpRequest);
+
+            string BuildErrorMessage(string httpResponseBody)
+            {
+                return
+                    $"{httpRequest.Method}: {httpRequest.RequestUri} failed with error code {httpResponse.StatusCode} using bearer token {httpRequest.Headers.Authorization.Parameter}. Response body: {httpResponseBody}";
```
muzukashiine: `httpRequest.Headers.Authorization.Parameter` will throw a null pointer exception since we're not checking that it exists.

Which begs the question: Should we check in the `SwedbankPayHttpClient`'s constructor if an authorization token exists? Should we set it in the constructor ourselves? Should we still set it in the `HttpClient` which is passed into the constructor like we're currently (trying) to do now?

My opinion is that we should keep this information as close to the `SwedbankPayHttpClient` somehow so that it's easier to use without having to remember all these details. I'm not a fan of modifying the `HttpClient` after the fact because it tends to lead to awkward code and/or race conditions as well, so there's that.

Let's discuss.

src/SwedbankPay.Sdk/SwedbankPayHttpClient.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// <typeparam name="TResponse"></typeparam>
+        /// <param name="httpMethod"></param>
+        /// <param name="url"></param>
+        /// <param name="payload"></param>
+        /// <exception cref="ArgumentNullException"></exception>
+        /// <exception cref="InvalidOperationException"></exception>
+        /// <exception cref="HttpRequestException"></exception>
+        /// <exception cref="HttpResponseException"></exception>
+        /// <returns></returns>
+        internal async Task<TResponse> SendHttpRequestAndProcessHttpResponse
```
muzukashiine: Do we really need this method?

src/SwedbankPay.Sdk/Exceptions/HttpResponseException.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Net.Http;
+
+namespace SwedbankPay.Sdk.Exceptions
+{
+    public class HttpResponseException : Exception
+    {
+        public HttpResponseException()
```
muzukashiine: Make private/delete.

src/SwedbankPay.Sdk/Exceptions/HttpResponseException.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+        }
+
+        public HttpResponseException(string message) : base(message)
```
muzukashiine: Make private/delete.

src/SwedbankPay.Sdk/Exceptions/HttpResponseException.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+        }
+
+        public HttpResponseException(string message, Exception innerException) : base(message, innerException)
```
muzukashiine: Make private/delete.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderRequestContainer.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace SwedbankPay.Sdk.PaymentOrders
+{
+    public class PaymentOrderRequestContainer
```
muzukashiine: What is the purpose of this class?

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderRequestContainer.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        public PaymentOrderRequestContainer(PaymentOrderRequest paymentOrder)
+        {
+            PaymentOrder = paymentOrder;
+        }
+
+
+        public PaymentOrderRequest PaymentOrder { get; }
```
muzukashiine: This is a `PaymentOrderRequest`. Please rename the property to avoid confusion.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderResponseContainer.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
+using System.Linq;
+
+using SwedbankPay.Sdk.Exceptions;
+
+namespace SwedbankPay.Sdk.PaymentOrders
+{
+    public class PaymentOrderResponseContainer
```
muzukashiine: What is the purpose of this class? Wouldn't these kind of methods belong on the response object itself?

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderResponseContainer.cs
```diff
@@ -0,0 +1,91 @@
...
+            PaymentOrder = paymentOrder;
+        }
+
+
+        public OperationList Operations { get; set; } = new OperationList();
+
+        public PaymentOrderResponse PaymentOrder { get; set; }
+
+
+        public Uri GetPaymentUrl()
```
muzukashiine: The other methods are quite similar to this, containing a lot of boilerplate code. Consider finding a way to merge them somehow.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderResponseContainer.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    public class PaymentOrderResponseContainer
+    { 
+        public PaymentOrderResponseContainer(PaymentOrderResponse paymentOrder)
+        {
+            PaymentOrder = paymentOrder;
+        }
+
+
+        public OperationList Operations { get; set; } = new OperationList();
```
muzukashiine: This operation list is never filled. Either way I'd recommend removing the setter, especially as we're initializing the list here.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderResponseContainer.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public PaymentOrderResponse PaymentOrder { get; set; }
```
muzukashiine: This property is a response object. Please rename to avoid confusion.

muzukashiine: Also remove the setter.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderRequestContainer.cs
karlfristedt76: To make the serialization match the json request

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderResponseContainer.cs
karlfristedt76: To make the serialization match the json request.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrder.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            foreach (var httpOperation in paymentOrderResponseContainer.Operations)
+            {
+                operations.Add(httpOperation.Rel, httpOperation);
+
+                switch (httpOperation.Rel.Value)
+                {
+                    case PaymentOrderResourceOperations.CreatePaymentOrderCapture:
+                        operations.Capture =
+                            new ExecuteRequestWrapper<TransactionRequestContainer, CaptureTransactionResponseContainer>(
```
muzukashiine: Consider finding a more readable way to do this. Essentially the code here is just Func<TRequest, TResponse>. What we currently have is a custom class with two different versions, using inheritance, and duplicating code already found in SwedbankPayHttpClient.

ExecuteWrapperBase has code to update payload and headers on a request. This code already exists when creating a request in SwedbankPayHttpClient and has been duplicated. I suggest we somehow enforce these headers when constructing the request, e.g. with the help of a static factory method. This would also mean the code in SwedbankPayHttpClient that is used to create requests needs to be deleted. Additionally, attaching the payload (and settings its content type) should also be extended onto the request class itself somehow so that it's not just loosely laying around like it is now. In conclusion, we don't really need ExecuteWrapperBase.

The two classes ExecuteRequestWrapper and ExecuteWrapper can be replaced by Func<TRequest, TResponse> and Func<TResponse> respectively. After all, what we want is a wrapper. If we're not doing anything significant in the wrapper we might as well use a simpler kind of wrapper. All we really want to do in the wrapper is to call the SwedbankPayHttpClient's "Send this request(request here)", a single line.

muzukashiine: Another thing to mention is that HttpOperation's request Accept header gets overwritten when it's put into an ExecuteWrapper, making the contentType argument of the constructor potentially useless.

zunkas: @muzukashiine we have talked to @asbjornu regarding this. Please discuss this with him a bit first.

README.md
```diff
@@ -2,26 +2,29 @@
...
 |-----------------------------|--------------|---------------------------------------------------|
-| AppVeyor                    | Windows      | [![Build status][appveyor-badge]][appveyor-build] |
-| Travis                      | Linux        | [![Build Status][travis-badge]][travis-build]     |
-| Azure DevOps                | Linux        | [![Build Status][azdo-badge]][azdo-build]         |
+| Azure DevOps                | Windows      | [![Build Status][azdo-badge]][azdo-build]         |
 
 ## About
 
-`PayEx.Client` is a `netstandard2` library to talk to PayEx direct REST APIs.
+`SwedbankPay.Sdk` is a `netstandard2` library to talk to SwedbankPay direct REST APIs.
```
asbjornu:
```suggestion
`SwedbankPay.Sdk` is a `netstandard2` library that allows you to interact with to Swedbank Pay's API Platform in a statically typed client.
```

README.md
```diff
@@ -2,26 +2,29 @@
...
 
 ## Supported APIs
 
-* Vipps
-  * create payment
-  * Vipps authorize
+* PaymentOrder
```
asbjornu:
```suggestion
* **Payment Order**
```

README.md
```diff
@@ -2,26 +2,29 @@
...
+  * create payment order
+  * get payment order
   * capture
   * cancel
   * reversal
+  * abort
+* Swish
```
asbjornu:
```suggestion
* **Swish Payments**
```

README.md
```diff
@@ -2,26 +2,29 @@
...
+  * create swish payment
+  * get swish payment
+  * abort
+  * reversal
 * CreditCard
```
asbjornu:
```suggestion
* **Card Payments**
```

PR #8, opened 2020-01-31
A total refactor of many files, excluding up to a dozen not yet touched

NullableInt:
Just some small clean ups to remove all warnings in Visual Studio and reduce the amount of messages Visual Studio outputs.

Opening this as a draft for now.


src/Samples/Sample.AspNetCore.SystemTests/PageObjectModels/Base/ControlAutomationDefinition.cs
```diff
@@ -0,0 +1,27 @@
+﻿
+using Atata;
+
+namespace Sample.AspNetCore.SystemTests.PageObjectModels.Base
```
asbjornu: Please leave the Samples project alone. :)

src/SwedbankPay.Sdk.Tests/Json/CustomAmountConverterTests.cs
```diff
@@ -19,13 +19,13 @@ public void CanDeSerialize_Amount()
         {
             //ARRANGE
 
-            var jsonObject = new JObject { { "xX123xxaddress", this.expectedAmount } };
+            var jsonObject = new JObject { { "xX123xxaddress", expectedAmount } };
```
asbjornu: Please prefix fields with `this.`. :)

```suggestion
            var jsonObject = new JObject { { "xX123xxaddress", this.expectedAmount } };
```

src/SwedbankPay.Sdk.Tests/TestHelpers/MockFileProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        private readonly IEnumerable<IFileInfo> _files;
+        private readonly Dictionary<string, IChangeToken> _changeTokens;
+
+        public MockFileProvider()
+        { }
+
+        public MockFileProvider(params IFileInfo[] files)
+        {
+            _files = files;
```
asbjornu:
```suggestion
            this.files = files;
```

src/SwedbankPay.Sdk.Tests/TestHelpers/MockFileProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Linq;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Primitives;
+using Moq;
+
+namespace SwedbankPay.Sdk.Tests.TestHelpers
+{
+    public class MockFileProvider : IFileProvider
+    {
+        private readonly IEnumerable<IFileInfo> _files;
```
asbjornu:
```suggestion
        private readonly IEnumerable<IFileInfo> files;
```

src/SwedbankPay.Sdk.Tests/TestHelpers/MockFileProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        private readonly Dictionary<string, IChangeToken> _changeTokens;
```
asbjornu:
```suggestion
        private readonly Dictionary<string, IChangeToken> changeTokens;
```

src/SwedbankPay.Sdk.Tests/TestHelpers/MockFileProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        { }
+
+        public MockFileProvider(params IFileInfo[] files)
+        {
+            _files = files;
+        }
+
+        public MockFileProvider(params KeyValuePair<string, IChangeToken>[] changeTokens)
+        {
+            _changeTokens = changeTokens.ToDictionary(
```
asbjornu:
```suggestion
            this.changeTokens = changeTokens.ToDictionary(
```

src/SwedbankPay.Sdk/Payments/Swish/Payment.cs
```diff
@@ -18,13 +20,12 @@ private Payment(PaymentResponse paymentResponse, SwedbankPayHttpClient client)
                 {
                     case PaymentResourceOperations.UpdatePaymentAbort:
                         operations.Abort = async () =>
-                            await client.SendHttpRequestAndProcessHttpResponse<PaymentResponse>(
-                                httpOperation.Request.AttachPayload(new PaymentAbortRequest()));
+                            await client.PatchAsJsonAsync<PaymentResponse>(httpOperation.Href, new PaymentAbortRequest());
```
zunkas: This does not follow the operations design. Here we assume it's a Patch again instead of using the Method from the Operations response.

src/SwedbankPay.Sdk/Payments/Card/Payment.cs
```diff
@@ -30,26 +32,22 @@ private Payment(PaymentResponse paymentResponse, SwedbankPayHttpClient client)
 
                     case PaymentResourceOperations.DirectAuthorization:
                         operations.DirectAuthorization = async payload =>
-                            await client.SendHttpRequestAndProcessHttpResponse<AuthorizationResponse>(
-                                httpOperation.Request.AttachPayload(payload));
+                            await client.PostAsJsonAsync<AuthorizationResponse>(httpOperation.Href, payload);
```
zunkas: This does not follow the operations design. Here we assume it's a Post instead of using the Method from the Operations response.

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrder.cs
```diff
@@ -21,19 +22,19 @@ public class PaymentOrder
...
+                        operations.Cancel = async payload => await client.PostAsJsonAsync<CancellationResponse>(httpOperation.Href, payload);
                         break;
                     case PaymentOrderResourceOperations.CreatePaymentOrderReversal:
-                        operations.Reversal = async payload => await client.SendHttpRequestAndProcessHttpResponse<ReversalResponse>(httpOperation.Request.AttachPayload(payload));
+                        operations.Reversal = async payload => await client.PostAsJsonAsync<ReversalResponse>(httpOperation.Href, payload);
                         break;
                     case PaymentOrderResourceOperations.UpdatePaymentOrderUpdateOrder:
-                        operations.Update = async payload => await client.SendHttpRequestAndProcessHttpResponse<PaymentOrderResponse>(httpOperation.Request.AttachPayload(payload));
+                        operations.Update = async payload => await client.PatchAsJsonAsync<PaymentOrderResponse>(httpOperation.Href, payload);
```
zunkas: This does not follow the operations design. Here we assume it's a Patch instead of using the Method from the Operations response.

src/SwedbankPay.Sdk/SwedbankPayClient.cs
```diff
@@ -13,8 +10,7 @@ namespace SwedbankPay.Sdk
 {
     public class SwedbankPayClient : ISwedbankPayClient
     {
-        public SwedbankPayClient(HttpClient httpClient,
-                                 ILogger logger = null)
+        public SwedbankPayClient(HttpClient httpClient)
```
zunkas: ILogger should be passed so we are able to log errors. With this there is no logging in the SDK

src/SwedbankPay.Sdk.Tests/PaymentOrderTests.cs
```diff
@@ -97,7 +85,7 @@ public async Task CreateAndUpdatePaymentOrder_ShouldReturnPaymentOrderWithNewAmo
             var updateRequest = new UpdateRequest(Amount.FromDecimal(newAmount), Amount.FromDecimal(newVatAmount));
             Assert.NotNull(paymentOrder.Operations.Update);
 
-            var response = await paymentOrder.Operations.Update?.Invoke(updateRequest);
+            var response = await paymentOrder.Operations.Update(updateRequest);
```
zunkas: Update could be null here. So that's why ?.Invoke() was used. 
There should also be an Assert.NotNull(response) in case it would be null.

src/SwedbankPay.Sdk.Tests/PaymentTests/SwishPaymentTests.cs
```diff
@@ -22,8 +22,9 @@ public async Task AbortPayment()
             var payment = await this.Sut.Payment.CreateSwishPayment(paymentRequest);
 
             Assert.NotNull(payment);
+            Assert.NotNull(payment.Operations.Abort);
 
-            var paymentResponseContainer = await payment.Operations.Abort.Invoke();
+            var paymentResponseContainer = await payment.Operations.Abort();
```
zunkas: Abort could be null. Nullcheck should still exist

src/SwedbankPay.Sdk.Tests/TestHelpers/MemoryFileInfo.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.IO;
+using Microsoft.Extensions.FileProviders;
+
+namespace SwedbankPay.Sdk.Tests.TestHelpers
+{
+    public class MemoryFileInfo : IFileInfo
```
zunkas: What is this class used for?

src/SwedbankPay.Sdk.Tests/TestHelpers/MockFileProvider.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Primitives;
+using Moq;
+
+namespace SwedbankPay.Sdk.Tests.TestHelpers
+{
+    public class MockFileProvider : IFileProvider
```
zunkas: What is this class used for?

src/SwedbankPay.Sdk.Tests/TestHelpers/MemoryFileInfo.cs
zunkas: I noticed it was there earlier also. But i couldn't see that either this or MockFileProvider is used.

NullableInt: Good catch. It seems this file is no longer in use.

src/SwedbankPay.Sdk.Tests/TestHelpers/MockFileProvider.cs
NullableInt: Good catch. It seems this file is no longer in use.

src/SwedbankPay.Sdk/SwedbankPayClient.cs
NullableInt: The `ILogger` were removed as there were no reference for it.

src/SwedbankPay.Sdk.Tests/PaymentOrderTests.cs
NullableInt: If it can be null then the `Assert.NotNull(paymentOrder.Operations.Update)` two lines above should fail.

src/SwedbankPay.Sdk/SwedbankPayClient.cs
zunkas: I understand, it was used in SwedbankPayHttpClient to log errors. Check with @asbjornu regarding this.

src/SwedbankPay.Sdk.Tests/PaymentOrderTests.cs
zunkas: True

src/SwedbankPay.Sdk.Tests/PaymentTests/SwishPaymentTests.cs
NullableInt: If it is null then the `Assert.NotNull(payment.Operations.Abort);` should fail.

Is the usage of the operations not to be like this:
```
var paymentOrder = await _swedbankPayClient.PaymentOrder.Get(paymentOrderUri));
if (paymentOrder.Operations.Abort != null)
{
    await paymentOrder.Operations.Abort();
}
```

src/SwedbankPay.Sdk/Payments/Swish/Payment.cs
NullableInt: I see.
I'll find a way to introduce that back in the code.

src/SwedbankPay.Sdk.Tests/PaymentTests/SwishPaymentTests.cs
zunkas: Sorry, i didn't notice you added `Assert.NotNull(payment.Operations.Abort);` 
In this case Abort could not be null then.

You could do it like above, or just use the null-conditional operator and check if response is null.

src/SwedbankPay.Sdk/SwedbankPayClient.cs
NullableInt: True true. I'll change up the client so it once again logs.

src/SwedbankPay.Sdk.Tests/TestHelpers/MockFileProvider.cs
zunkas: Actually it was in use i realized now when the build failed, but in Release mode. Switch to Release and check TestHelper class. There is also a missing using statement for Release mode i think.

src/SwedbankPay.Sdk/SwedbankPayClient.cs
asbjornu: We need the logging, but we should also investigate whether the two different client classes can be merged into one.

NullableInt:
Okay, I just did a test to simplify the experience when using the SDK. I'm very open for discussions on this.

The thought was that when getting a `ISwedbankPayClient` and accessing its properties it feels a bit weird as a developer.
Having to access the client and then choose if I want payments, consumers, or paymentOrders felt a bit odd.
Instead I thought it would be good if the users of the SDK had to make a decision when choosing what type of client they want.
As such a user have to choose if they want a `IPaymentOrdersClient`, `IConsumersClient` or `IPaymentsClient`. Or if they want to have access to it all they can use `ISwedbankPayClient`.

I'm not sure about this change and I'm looking for feedback and opinions, but the end goal is to reduce the amount of information the end user have to process when using the client(s).

For example:

```
public class MyController: ApiController
{
    public MyController(IPaymentOrdersClient paymentOrderClient, IConsumersClient consumersClient)
   ....

   //In some method
   consumersClient.Initiate(stuff here);
  

  // In a different method
    paymentOrderClient.Create(paymentOrder
}
```

This is just to reduce the amount of information and options the end user have when they access their client.

src/SwedbankPay.Sdk/SwedbankPayClient.cs
NullableInt: Why do we need the logging?

NullableInt:
`enum`s and the deserialization of them should support `Unknown` so that when new types are added in the API (such as now with `PriceType`) it won't crash the SDK.

src/Samples/Sample.AspNetCore.SystemTests/Test/PaymentTests/Payment/PaymentSaleTests.cs
```diff
@@ -30,7 +30,7 @@ public async Task Payment_Swish_Sale(Product[] products, PayexInfo payexInfo)
                 .Actions.Rows[y => y.Name.Value.Contains(PaymentResourceOperations.ViewPayment)].Should.BeVisible()
                 .Actions.Rows.Count.Should.Equal(3);
 
-            var swishPayment = await SwedbankPayClient.Payment.GetSwishPayment(paymentLink, PaymentExpand.All);
+            var swishPayment = await SwedbankPayClient.GetSwishPayment(paymentLink, PaymentExpand.All);
```
asbjornu: Is it possible to have this read like the following?

```suggestion
            var swishPayment = await SwedbankPayClient.SwishPayments.Get(paymentLink, PaymentExpand.All);
```

src/Samples/Sample.AspNetCore.SystemTests/Test/PaymentTests/PaymentOrder/Abort/AbortTests.cs
```diff
@@ -26,7 +26,7 @@ public async Task Abort_PaymentOrder(Product[] products)
 
             var orderLink = message.OriginalString.Substring(message.OriginalString.IndexOf("/")).Replace(" has been Aborted", "");
 
-            var order = await SwedbankPayClient.PaymentOrder.Get(new Uri(orderLink, UriKind.RelativeOrAbsolute), SwedbankPay.Sdk.PaymentOrders.PaymentOrderExpand.All);
+            var order = await SwedbankPayClient.Get(new Uri(orderLink, UriKind.RelativeOrAbsolute), SwedbankPay.Sdk.PaymentOrders.PaymentOrderExpand.All);
```
asbjornu: Is it possible to have this read like the following?

```suggestion
            var order = await SwedbankPayClient.PaymentOrders.Get(new Uri(orderLink, UriKind.RelativeOrAbsolute), SwedbankPay.Sdk.PaymentOrders.PaymentOrderExpand.All);
```

src/Samples/Sample.AspNetCore/Controllers/CheckOutController.cs
```diff
@@ -104,7 +104,7 @@ public async Task<SwedbankPay.Sdk.Payments.Card.Payment> CreateCardPayment()
                                                                                    new PayeeInfo(this.payeeInfoOptions.PayeeId,
                                                                                                  this.payeeInfoOptions.PayeeReference));
 
-                SwedbankPay.Sdk.Payments.Card.Payment cardPayment = await this.swedbankPayClient.Payment.CreateCreditCardPayment(cardRequest);
+                SwedbankPay.Sdk.Payments.Card.Payment cardPayment = await this.swedbankPayClient.CreateCreditCardPayment(cardRequest);
```
asbjornu: Can we have this read like the following? 

```suggestion
                SwedbankPay.Sdk.CardPayments.CardPayment cardPayment = await this.swedbankPayClient.CardPayments.Create(cardRequest);
```

NullableInt: I've ripped out that part of the `SwedbankPayClient`. If you want to use one singular part of the api at a time you could just use `IPaymentsClient` instead of the `ISwedbankPayClient`.

src/Samples/Sample.AspNetCore.SystemTests/Test/PaymentTests/PaymentOrder/Abort/AbortTests.cs
NullableInt: I've ripped out that part of the `SwedbankPayClient`. If you want to use one singular part of the api at a time you could just use `IPaymentOrdersClient` instead of the `ISwedbankPayClient`.

src/Samples/Sample.AspNetCore.SystemTests/Test/PaymentTests/Payment/PaymentSaleTests.cs
NullableInt: I've ripped out that part of the `SwedbankPayClient`. If you want to use one singular part of the api at a time you could just use `IPaymentsClient` instead of the `ISwedbankPayClient`.
But I can see what I can do.

src/Samples/Sample.AspNetCore/Controllers/PaymentController.cs
```diff
@@ -253,21 +254,21 @@ public async Task<IActionResult> Reversal(string paymentId, Instrument instrumen
         }
 
 
-        private async Task<CaptureRequest> GetCaptureRequest(string description)
+        private async Task<SwedbankPay.Sdk.PaymentOrders.CaptureRequest> GetCaptureRequest(string description)
```
asbjornu:
```suggestion
        private async Task<PaymentOrderCaptureRequest> GetCaptureRequest(string description)
```

src/SwedbankPay.Sdk/Payments/SwishPayments/Operations.cs
```diff
@@ -2,14 +2,14 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 
-namespace SwedbankPay.Sdk.Payments.Swish
+namespace SwedbankPay.Sdk.Payments.SwishPayments
 {
     public class Operations : Dictionary<LinkRelation, HttpOperation>
     {
         public Func<Task<PaymentResponse>> Abort { get; internal set; }
         public Func<ReversalRequest, Task<ReversalResponse>> CreateReversal { get; internal set; }
```
asbjornu: It's OK that the `const string` is named `CreateReversal` to reflect the underlying `rel` value, but the logical function performed is `Reverse` so it makes more sense to expose it as `Reverse` in C#. The same goes for `CreateCapture` (should be `Capture`), `CreateCancel` (should be `Cancel`), etc.

```suggestion
        public Func<ReversalRequest, Task<ReversalResponse>> Reverse { get; internal set; }
```

src/SwedbankPay.Sdk/Payments/SwishPayments/Operations.cs
```diff
@@ -2,14 +2,14 @@
...
         public Func<SaleRequest, Task<SaleResponse>> CreateSale { get; internal set; }
-        public HttpOperation this[LinkRelation rel] => ContainsKey(rel) ? base[rel] : null;
+        public new HttpOperation this[LinkRelation rel] => ContainsKey(rel) ? base[rel] : null;
```
asbjornu: `new` is a bit dirty. Can we just implement `IDictionary` instead of inheriting `Dictionary`?

src/SwedbankPay.Sdk/Payments/SwishPayments/PaymentRequestObject.cs
```diff
@@ -0,0 +1,56 @@
...
+                                                List<Price> prices,
+                                                string description,
+                                                string payerReference,
+                                                string userAgent,
+                                                CultureInfo language,
+                                                Urls urls,
+                                                PayeeInfo payeeInfo,
+                                                PrefillInfo prefillInfo,
+                                                SwishRequest swishRequest,
+                                                Dictionary<string, object> metaData = null)
```
asbjornu: Arguments with `null` default value can just be removed from the constructor and have their property changed to a `public` setter instead. Only required, non-default arguments should exist in the constructor.

src/SwedbankPay.Sdk/Payments/SwishPayments/PaymentResponse.cs
```diff
@@ -3,17 +3,17 @@
 
 using SwedbankPay.Sdk.PaymentOrders;
 
-namespace SwedbankPay.Sdk.Payments.Swish
+namespace SwedbankPay.Sdk.Payments.SwishPayments
 {
     public class PaymentResponse
```
asbjornu:
```suggestion
    public class SwishPaymentResponse
```

src/SwedbankPay.Sdk/Payments/SwishPayments/ReversalRequest.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace SwedbankPay.Sdk.Payments.SwishPayments
+{
+    public class ReversalRequest
```
asbjornu:
```suggestion
    public class SwishReversalRequest
```

src/SwedbankPay.Sdk/Payments/SwishPayments/ReversalTransaction.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace SwedbankPay.Sdk.Payments.SwishPayments
+{
+    public class ReversalTransaction
```
asbjornu:
```suggestion
    public class SwishReversalTransaction
```

src/SwedbankPay.Sdk/Payments/SwishPayments/Operations.cs
NullableInt: Sure, implementing `IDictionary` is 15 methods.

src/SwedbankPay.Sdk/Payments/SwishPayments/PaymentRequestObject.cs
zunkas: Isn't this against the Immutable principle?

NullableInt: I agree with zunkas. Plus having to mix both makes list initialization annoying. Either all as properties or all as constructor, don't mix both if we don't have to.

asbjornu: The way it is now we end up with enormous constructors, which is less than ideal. I know @muzukashiine had issues with this as well. The greatest value of constructors is being able to validate the arguments to ensure that they follow our business rules of not being `null`, having invalid data, etc., before an instance of a class can be constructed. Once constructed, that instance should follow the business rules. If a class has a lot of optional fields, having them as constructor arguments provide no real value, as there is no validation that can be performed on them.

src/SwedbankPay.Sdk/Payments/SwishPayments/Operations.cs
asbjornu: Please add a base class that implements the index accessor and `IDictionary` so we can get rid of the `new`.

NullableInt: Can do

src/Samples/Sample.AspNetCore/Controllers/CheckOutController.cs
```diff
@@ -84,13 +80,13 @@ public async Task<PaymentOrder> CreatePaymentOrder(string consumerProfileRef = n
             }
         }
 
-        public async Task<SwedbankPay.Sdk.Payments.Card.Payment> CreateCardPayment()
+        public async Task<SwedbankPay.Sdk.Payments.CardPayments.CardPayment> CreateCardPayment()
```
asbjornu: I suppose it's an easy task to remove all these superfluous namespaces with ReSharper once this PR is merged, @zunkas?

src/SwedbankPay.Sdk.Tests/TestBuilders/ConsumersRequestContainerBuilder.cs
```diff
@@ -8,18 +8,16 @@ namespace SwedbankPay.Sdk.Tests.TestBuilders
...
         private NationalIdentifier nationalIdentifier;
         private Operation operation;
         private IEnumerable<RegionInfo> shippingAddressRestrictedToCountryCodes;
 
 
         public ConsumersRequest Build()
         {
-            return new ConsumersRequest(this.language, this.shippingAddressRestrictedToCountryCodes, this.operation, this.msisdn,
-                                        this.email, this.nationalIdentifier);
+            return new ConsumersRequest(this.language, this.shippingAddressRestrictedToCountryCodes, this.operation, null,
```
asbjornu: Always sending `null` for `msisdn` and `email` seems like an unintentional "fix".

src/SwedbankPay.Sdk.Tests/TestHelpers/MemoryFileInfo.cs
asbjornu: If `MemoryFileInfo` is not in use, can't it be deleted?

src/SwedbankPay.Sdk/ISwedbankPayClient.cs
```diff
@@ -6,8 +6,8 @@ namespace SwedbankPay.Sdk
 {
     public interface ISwedbankPayClient
     {
-        IConsumersResource Consumers { get; }
-        IPaymentsResource Payment { get; }
-        IPaymentOrderResource PaymentOrder { get; }
+        public IPaymentOrdersResource PaymentOrders { get; }
```
zunkas: public is not a valid interface member declaration

src/SwedbankPay.Sdk/Payments/IPaymentsResource.cs
```diff
@@ -1,45 +1,11 @@
...
-
-
-        ///// <summary>
-        /////     Gets an existing swish payment.
-        ///// </summary>
-        ///// <param name="id"></param>
-        ///// <param name="paymentExpand"></param>
-        ///// <returns></returns>
-        Task<Swish.Payment> GetSwishPayment(Uri id, PaymentExpand paymentExpand = PaymentExpand.None);
+        public ICardPaymentsResource CardPayments { get; set; }
```
zunkas: remove public, and should this really have a setter?

src/SwedbankPay.Sdk/Extensions/HttpClientExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public static Task<T> PostAsJsonAsync<T>(this HttpClient httpClient, Uri uri, object payload)
+            where T : class
+        {
+            return httpClient.SendAndProcessAsync<T>(HttpMethod.Post, uri, payload);
+        }
+
+
+        public static Task<T> PutAsJsonAsync<P, T>(this HttpClient httpClient, Uri uri, object payload)
```
zunkas: What is P in <P, T> ?

src/SwedbankPay.Sdk/Extensions/HttpRequestMessageExtensions.cs
```diff
@@ -3,7 +3,7 @@
 
 using Newtonsoft.Json;
 
-namespace SwedbankPay.Sdk
+namespace SwedbankPay.Sdk.Extensions
```
zunkas: Doesn't seem to be used anymore. Remove?

src/Samples/Sample.AspNetCore/Controllers/CheckOutController.cs
zunkas: Move everything to Payments instead? 
Should be easy

src/SwedbankPay.Sdk/Payments/SwishPayments/Payment.cs
```diff
@@ -1,14 +1,16 @@
-﻿using System;
+﻿using SwedbankPay.Sdk.Extensions;
+using System;
+using System.Net.Http;
 using System.Threading.Tasks;
 
-namespace SwedbankPay.Sdk.Payments.Swish
+namespace SwedbankPay.Sdk.Payments.SwishPayments
 {
     public class Payment
```
zunkas: Should this also be renamed to SwishPayment?

src/SwedbankPay.Sdk.Tests/TestHelpers/MemoryFileInfo.cs
zunkas: It is used in TestHelper. If build in Release.

src/SwedbankPay.Sdk/Extensions/HttpRequestMessageExtensions.cs
NullableInt: I don't get this comment, would you kindly elaborate on this?

src/SwedbankPay.Sdk/Payments/SwishPayments/Payment.cs
NullableInt: That it should! Nice catch.

src/SwedbankPay.Sdk/Extensions/HttpClientExtensions.cs
NullableInt: A forgotten leftover from a previous experiment where the syntax used was:
`public static Task<T> PutAsJsonAsync<P, T>(this HttpClient httpClient, Uri uri, P payload)`, but this made using the method more cumbersome than it was worth.

src/SwedbankPay.Sdk/Extensions/HttpRequestMessageExtensions.cs
NullableInt: Ah, you meant the entire file, I'll remove it.

src/SwedbankPay.Sdk/SwedbankPay.Sdk.csproj
```diff
@@ -20,8 +20,10 @@
     </None>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.0.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.1" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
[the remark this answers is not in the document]
NullableInt: Reverted back to 11.0.1

src/SwedbankPay.Sdk.Tests/TestBuilders/ConsumersRequestContainerBuilder.cs
NullableInt: I removed them, is this fine?

src/SwedbankPay.Sdk/SwedbankPay.Sdk.csproj
```diff
@@ -20,7 +20,6 @@
     </None>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.1" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.1" />
```
zunkas: I think you should be able to remove this as well now?

src/SwedbankPay.Sdk/Payments/SwishPayments/SwishPayment.cs
```diff
@@ -1,14 +1,16 @@
...
+    public class SwishPayment
     {
-        private Payment(PaymentResponse paymentResponse, SwedbankPayHttpClient client)
+        private SwishPayment(SwishPaymentPaymentResponse paymentResponse, HttpClient client)
         {
             PaymentResponse = paymentResponse.Payment;
-            var operations = new Operations();
+            var operations = new SwishPaymentOperations();
 
             foreach (var httpOperation in paymentResponse.Operations)
```
asbjornu: This loop should be moved into the constructor of `SwishPaymentOperations` for [SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle) and immutability.

src/SwedbankPay.Sdk/RelationConstants.cs
```diff
@@ -16,38 +16,4 @@ public static class ConsumerResourceOperations
         public const string ViewConsumerIdentification = "view-consumer-identification";
         public const string RedirectConsumerIdentification = "redirect-consumer-identification";
     }
-
```
asbjornu: `RelationConstants.cs` should be renamed to `ConsumerResourceOperations.cs` and moved to a more appropriate place.

src/SwedbankPay.Sdk/ResourceBase.cs
```diff
@@ -1,16 +1,17 @@
...
 using System.Collections.Generic;
+using System.Net.Http;
 
 namespace SwedbankPay.Sdk
 {
-    internal abstract class ResourceBase
+    public abstract class ResourceBase
     {
-        internal SwedbankPayHttpClient swedbankPayHttpClient;
+        internal HttpClient httpClient;
```
asbjornu: Why does `HttpClient` need to be `internal`?

src/SwedbankPay.Sdk/ResourceBase.cs
```diff
@@ -1,16 +1,17 @@
...
 
 
-        protected ResourceBase(SwedbankPayHttpClient swedbankPayHttpClient)
+        protected ResourceBase(HttpClient httpClient)
         {
-            this.swedbankPayHttpClient = swedbankPayHttpClient;
+            this.httpClient = httpClient;
```
asbjornu:
```suggestion
            this.httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
```

src/Samples/Sample.AspNetCore.SystemTests/Test/PaymentTests/Base/PaymentTests.cs
```diff
@@ -51,7 +56,11 @@ public void OneTimeSetUp()
             httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", Environment.GetEnvironmentVariable("Payex.Api.Token", EnvironmentVariableTarget.User));
             #endif
 
-            SwedbankPayClient = new SwedbankPayClient(httpClient);
+            SwedbankPayClient = new SwedbankPayClient(
+                httpClient,
+                new PaymentOrdersResource(httpClient),
+                new ConsumersResource(httpClient),
+                new PaymentsResource(httpClient, new CardPaymentsResource(httpClient), new SwishPaymentsResource(httpClient)));
```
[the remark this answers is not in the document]
NullableInt: Done

src/SwedbankPay.Sdk/RelationConstants.cs
NullableInt: Done

src/SwedbankPay.Sdk/Payments/SwishPayments/PaymentRequestObject.cs
muzukashiine: My opinion is that a bunch of arguments should not be passed to the constructor, only to be continually passed down. Group arguments into objects as early as possible and pass in these instead. Example: A complex `Person` class would not have 100 arguments if there were 100 details one would want to describe about a person. These would come grouped into objects that would themselves be grouped again and again and so on. My rule of thumb is if an argument list grows past X elements I need to think about how I can group them better.

Having both constructor arguments and public setters uses two different styles. I would prefer we stick to one, like mentioned before by NullableInt. However, I also agree that passing in optional arguments to the constructor can feel bloaty and unnecessary. But I'm not a fan of using public setters either, I much rather we use stricter setters (aka methods with clear business intent). That's what I would have done.

src/SwedbankPay.Sdk/Extensions/HttpClientExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+                    throw new HttpResponseException(
+                        httpResponseMessage,
+                        JsonConvert.DeserializeObject<ProblemResponse>(httpResponseBody));
+                }
+
+                return JsonConvert.DeserializeObject<T>(httpResponseBody, JsonSerialization.JsonSerialization.Settings);
+            }
+            catch (HttpResponseException ex)
+            {
+                httpResponseBody = await httpResponseMessage.Content.ReadAsStringAsync();
```
asbjornu:
```suggestion
```

src/SwedbankPay.Sdk/Payments/CardPayments/CardPaymentOperations.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public Func<CardPaymentCancelRequest, Task<CancellationResponse>> Cancel { get; }
+        public Func<CardPaymentCaptureRequest, Task<CaptureResponse>> Capture { get; }
+        public Func<CardPaymentAuthorizationRequest, Task<CardPaymentAuthorizationResponse>> DirectAuthorization { get; }
+        public Func<CardPaymentReversalRequest, Task<ReversalResponse>> Reverse { get; }
+        public HttpOperation DirectVerification { get; internal set; }
+        public HttpOperation PaidPayment { get; internal set; }
+        public HttpOperation RedirectAuthorization { get; internal set; }
+        public HttpOperation RedirectVerification { get; internal set; }
+        public HttpOperation Update { get; internal set; }
```
asbjornu: Why isn't the `Update` operation a `Func`?

src/SwedbankPay.Sdk/SwedbankPayClient.cs
```diff
@@ -1,44 +1,60 @@
...
+            if (this.httpClient.DefaultRequestHeaders?.Authorization?.Parameter == null)
+                throw new ArgumentException($"Please configure the {nameof(httpClient)} with an Authorization header.");
 
-        public IPaymentsResource Payment { get; }
+            var cardPaymentsResource = new CardPaymentsResource(this.httpClient);
+            var swishPaymentsResource = new SwishPaymentsResource(this.httpClient);
+
+            PaymentOrders = new PaymentOrdersResource(this.httpClient);
+            Consumers = new ConsumersResource(this.httpClient);
+            Payments = new PaymentsResource(this.httpClient, cardPaymentsResource, swishPaymentsResource);
```
asbjornu: Can `PaymentsResource` provide a constructor that initializes `CardPaymentsResource` and `SwishPaymentsResource` itself?

src/SwedbankPay.Sdk/Extensions/HttpClientExtensions.cs
NullableInt: Can't. This will create a error code `CS0165` on line 59

src/SwedbankPay.Sdk/Payments/CardPayments/CardPaymentOperations.cs
NullableInt: No idea, I can look into it.

src/SwedbankPay.Sdk/Consumers/ConsumerOperations.cs
```diff
@@ -0,0 +1,37 @@
...
+            }
+        }
+
+
+        protected ConsumerOperations(
+            SerializationInfo info)
+            : base()
+        {
+            RedirectConsumerIdentification = (HttpOperation)info.GetValue("RedirectConsumerIdentification", typeof(HttpOperation));
+            RedirectConsumerIdentification = (HttpOperation)info.GetValue("ViewConsumerIdentification", typeof(HttpOperation));
```
asbjornu:
```suggestion
            ViewConsumerIdentification = (HttpOperation)info.GetValue("ViewConsumerIdentification", typeof(HttpOperation));
```

src/SwedbankPay.Sdk/LinkRelation.cs
```diff
@@ -57,7 +59,7 @@ public static readonly LinkRelation
         public static readonly LinkRelation PaidPayment = new LinkRelation(nameof(PaidPayment), PaymentResourceOperations.PaidPayment);
 
         public static readonly LinkRelation UpdatePaymentAbort =
-            new LinkRelation(nameof(UpdatePaymentAbort), PaymentResourceOperations.UpdatePaymentAbort);
+            new LinkRelation(nameof(UpdatePaymentAbort), PaymentResourceOperations.Abort);
```
asbjornu: For consistency I think `UpdatePaymentAbort` is correct at this level.

```suggestion
            new LinkRelation(nameof(UpdatePaymentAbort), PaymentResourceOperations.UpdatePaymentAbort);
```

src/SwedbankPay.Sdk/PaymentResourceOperations.cs
```diff
@@ -11,28 +11,9 @@
...
-        public const string CreatePaymentOrderCancel = "create-paymentorder-cancel";
-        public const string RedirectVerification = "redirect-verification";
-        public const string CreatePaymentOrderReversal = "create-paymentorder-reversal";
-        public const string PaidPaymentOrder = "paid-paymentorder";
-    }
-
     public static class PaymentResourceOperations
     {
-        public const string UpdatePaymentAbort = "update-payment-abort";
+        public const string Abort = "update-payment-abort";
```
asbjornu:
```suggestion
        public const string UpdatePaymentAbort = "update-payment-abort";
```

src/SwedbankPay.Sdk/Payments/CardPayments/CardPaymentOperations.cs
```diff
@@ -0,0 +1,81 @@
...
+    public class CardPaymentOperations : OperationsBase
+    {
+        public CardPaymentOperations(OperationList operations, HttpClient client)
+        {
+            foreach (var httpOperation in operations)
+            {
+                switch (httpOperation.Rel.Value)
+                {
+                    case PaymentResourceOperations.Abort:
+                        Update = async payload =>
```
asbjornu:
```suggestion
                        Abort = async payload => 
```

src/SwedbankPay.Sdk/Payments/CardPayments/CardPaymentOperations.cs
```diff
@@ -0,0 +1,81 @@
...
+                }
+                this.Add(httpOperation.Rel, httpOperation);
+            }
+        }
+
+        public Func<CardPaymentCancelRequest, Task<CancellationResponse>> Cancel { get; }
+        public Func<CardPaymentCaptureRequest, Task<CaptureResponse>> Capture { get; }
+        public Func<CardPaymentAuthorizationRequest, Task<CardPaymentAuthorizationResponse>> DirectAuthorization { get; }
+        public Func<CardPaymentReversalRequest, Task<ReversalResponse>> Reverse { get; }
+        public Func<CardPaymentAbortRequest, Task<CardPaymentResponse>> Update { get; }
```
asbjornu:
```suggestion
        public Func<CardPaymentAbortRequest, Task<CardPaymentResponse>> Abort { get; }
```

src/SwedbankPay.Sdk/Payments/PaymentsResource.cs
```diff
@@ -1,46 +1,26 @@
...
-
-
-        public async Task<Payment> CreateCreditCardPayment(PaymentRequest paymentRequest, PaymentExpand paymentExpand = PaymentExpand.None)
-        {
-            return await Payment.Create(paymentRequest, this.swedbankPayHttpClient, GetExpandQueryString(paymentExpand));
-        }
-
-
-        public async Task<Swish.Payment> GetSwishPayment(Uri id, PaymentExpand paymentExpand = PaymentExpand.None)
+        public ICardPaymentsResource CardPayments { get; private set ; }
```
asbjornu:
```suggestion
        public ICardPaymentsResource CardPayments { get; }
```

src/SwedbankPay.Sdk/Payments/PaymentsResource.cs
```diff
@@ -1,46 +1,26 @@
...
+        public ISwishPaymentsResource SwishPayments { get; private set; }
```
asbjornu:
```suggestion
        public ISwishPaymentsResource SwishPayments { get; }
```

src/SwedbankPay.Sdk/Payments/PaymentsResource.cs
```diff
@@ -1,46 +1,26 @@
...
+        
+        public PaymentsResource(HttpClient httpClient) : base(httpClient)
```
asbjornu:
```suggestion
        public PaymentsResource(HttpClient httpClient)
            : this(httpClient, new CardPaymentsResource(httpClient), new SwishPaymentsResource(httpClient))
```

src/SwedbankPay.Sdk/Payments/PaymentsResource.cs
```diff
@@ -1,46 +1,26 @@
...
         {
-            if (id == null)
-                throw new ArgumentNullException(nameof(id));
-
-            return await Swish.Payment.Get(id, this.swedbankPayHttpClient, GetExpandQueryString(paymentExpand));
+            this.CardPayments = new CardPaymentsResource(httpClient);
+            this.SwishPayments = new SwishPaymentsResource(httpClient);
```
asbjornu: With the constructor calling `this()` we can remove these two lines.

src/SwedbankPay.Sdk/Payments/PaymentsResource.cs
```diff
@@ -1,46 +1,26 @@
...
         }
 
-
-        public async Task<Swish.Payment> CreateSwishPayment(Swish.PaymentRequest paymentRequest,
-                                                            PaymentExpand paymentExpand = PaymentExpand.None)
+        public PaymentsResource(HttpClient httpClient, ICardPaymentsResource cardPaymentsResource, ISwishPaymentsResource swishPaymentsResource) : base(httpClient)
         {
-            return await Swish.Payment.Create(paymentRequest, this.swedbankPayHttpClient, GetExpandQueryString(paymentExpand));
+            this.CardPayments = cardPaymentsResource;
```
asbjornu: It seems like there are quite a few missing [guard clauses](https://deviq.com/guard-clause/) in our constructors. Can you please go over all constructors and ensure that all arguments are not null, not empty strings and similar so we don't bomb with `NullReferenceException` and similar after construction?

```suggestion
            this.CardPayments = cardPaymentsResource ?? throw new ArgumentNullException(nameof(cardPaymentsResource));
```

src/SwedbankPay.Sdk/Payments/PaymentsResource.cs
NullableInt: Doing this will give me `Error	CS7036	There is no argument given that corresponds to the required formal parameter 'httpClient' of 'ResourceBase.ResourceBase(HttpClient)'`

src/SwedbankPay.Sdk/Payments/PaymentsResource.cs
NullableInt: You'll have to show me what you mean here.

src/SwedbankPay.Sdk/Payments/SwishPayments/SwishPaymentOperations.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public SwishPaymentOperations(OperationList operations, HttpClient client)
+        {
+            foreach (var httpOperation in operations)
+            {
+                switch (httpOperation.Rel.Value)
+                {
+                    case PaymentResourceOperations.UpdatePaymentAbort:
+                        Abort = async () =>
+                            await client.SendAsJsonAsync<SwishPaymentPaymentResponse>(httpOperation.Method, httpOperation.Href, new PaymentAbortRequest());
```
asbjornu: Shouldn't `PaymentAbortRequest` be passed in as a `payload` argument?

```suggestion
                            await client.SendAsJsonAsync<SwishPaymentPaymentResponse>(httpOperation.Method, httpOperation.Href, payload);
```

src/SwedbankPay.Sdk/Payments/CardPayments/CardPaymentAuthorizationRequestTransaction.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace SwedbankPay.Sdk.Payments.CardPayments
+{
+    public class CardPaymentAuthorizationRequestTransaction
+    {
+        protected internal CardPaymentAuthorizationRequestTransaction(string cardNumber,
```
zunkas: This class needs [JsonConstructor], it won't deserialize otherwise since it's protected internal.

NullableInt: Fixed 👍

src/Samples/Sample.AspNetCore/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -20,20 +29,23 @@ public static class ServiceCollectionExtensions
             services.Configure<SwedbankPayConnectionSettings>(swedbankPayConSettings);
 
             var swedBankPayOptions = swedbankPayConSettings.Get<SwedbankPayConnectionSettings>();
-            swedBankPayOptions.Token = configuration["Token"];
+            swedBankPayOptions.Token = "6430eee8fe8a2902f74494087ae79b9ff2e2f9d1efd3eeac1110e79d3502df24";
```
asbjornu: Is really hard coding a token here a good idea?

src/Samples/Sample.AspNetCore/appsettings.json
```diff
@@ -9,10 +9,10 @@
   "AllowedHosts": "*",
 
   "PayeeInfo": {
-    "PayeeId": "91a4c8e0-72ac-425c-a687-856706f9e9a1"
+    "PayeeId": "c60d3e0b-bad8-40fb-b02b-af75fca9702c"
```
asbjornu: Wouldn't it be better to have a `PayeeId` with the value of `Guid.Empty` and do a check for that on `Startup` so we can fail with a user friendly error message that the Sample project needs to be properly configured as per the `README`?

src/Samples/Sample.AspNetCore/appsettings.json
```diff
@@ -9,10 +9,10 @@
...
   },
   "SwedbankPayConnectionSettings": {
-    "ApiBaseUrl": "https://api.externalintegration.payex.com"
+    "ApiBaseUrl": "https://api.stage.payex.com"
```
asbjornu: Why the change to `stage`?

src/SwedbankPay.Sdk/Payments/SwishPayments/SwishPaymentOperations.cs
```diff
@@ -0,0 +1,51 @@
...
+                        break;
+                }
+                this.Add(httpOperation.Rel, httpOperation);
+            }
+        }
+
+        public Func<PaymentAbortRequest, Task<SwishPaymentPaymentResponse>> Abort { get; }
+        public Func<SwishPaymentReversalRequest, Task<ReversalResponse>> Reverse { get; }
+        public Func<SwishPaymentSaleRequest, Task<SwishPaymentSaleResponse>> Sale { get; }
+        public HttpOperation PaidPayment { get; internal set; }
```
asbjornu:
```suggestion
        public HttpOperation PaidPayment { get; }
```

src/SwedbankPay.Sdk/Payments/SwishPayments/SwishPaymentOperations.cs
```diff
@@ -0,0 +1,51 @@
...
+        public HttpOperation RedirectSale { get; internal set; }
```
asbjornu:
```suggestion
        public HttpOperation RedirectSale { get; }
```

src/SwedbankPay.Sdk/Payments/SwishPayments/SwishPaymentOperations.cs
```diff
@@ -0,0 +1,51 @@
...
+        public HttpOperation ViewSales { get; internal set; }
```
asbjornu:
```suggestion
        public HttpOperation ViewSales { get; }
```

PR #10, opened 2020-02-11
Feature/dx 677 add vipps payments

trieutran91:


src/SwedbankPay.Sdk/Payments/VippsPayments/VippsPayment.cs
```diff
@@ -0,0 +1,45 @@
...
+            var operations = new VippsPaymentOperations(paymentResponse.Operations, client);
+            Operations = operations;
+        }
+
+        public VippsPaymentOperations Operations { get; }
+
+        public PaymentResponseObject PaymentResponse { get; }
+
+
+        internal static async Task<VippsPayment> Create(VippsPaymentRequest paymentRequest,
```
asbjornu: Hm, same here as in MobilePay. Is this how `Create` is implemented in the other payment instruments?

PR #11, opened 2020-02-11
Feature/dx 687 add invoice payments

trieutran91:


trieutran91:
I've requests should be matching with the documentation now. 
I'm still awaiting feedback on the `campaignCode` property when we create a payment.
It is documented here https://wiki.payex.com/xwiki/bin/view/Team%20Area/eCommerce/Domains%20%26%20Services/Payment1/Invoice%20Payments/Invoice%20Payment%20API/POST%20psp%2Finvoice%2Fpayments
that the POST request has that property. The question is whether it is in production or not.

PR #17, opened 2020-02-19
Add new operation to link relation list

NullableInt:
Having the SDK crash because of unknown operations is not maintainable, I'll create a issue for it and start looking into fixing it.


asbjornu:
@zunkas, can you please figure out why this PR fails the `sdk-dotnet-macOS-Dev-CI` test?

PR #23, opened 2020-02-21
Release/v1.1.0

NullableInt:
New payment methods:
MobilePay
Invoice
Vipps


NullableInt:
Trying to get a updated NuGet package that's not an alpha version

PR #24, opened 2020-02-27
Remove dependency on logging package

NullableInt:
Dependency is not needed, and is a blocker for people implementing this in .Net Core 2.2

### DIFF
--- a/src/SwedbankPay.Sdk/SwedbankPay.Sdk.csproj
+++ b/src/SwedbankPay.Sdk/SwedbankPay.Sdk.csproj
@@ -20,7 +20,6 @@
     </None>
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


NullableInt:
Issue found by @trieutran91

zunkas:
I thought of this also when implementing the SDK in the Episerver plugin. This package isn't used right now since the logging occurs with the extensions package, so it could be removed from here.
But right now we don't have any logging unless using the extension package which only can be used in .net core 3? What do you think about this @asbjornu ? I suppose there should be some logging for e.g. net framework also?

asbjornu:
As long as logging is possible to add externally, I see no reason why we should have this here. The fewer external dependencies in the SDK, the better.

NullableInt:
I'll add it to the extensions package.

PR #74, opened 2020-05-27
Changed style used in Paymentmenu iframe

karlfristedt76:


src/Samples/Sample.AspNetCore/Views/Checkout/_SwedbankPayCheckout.cshtml
```diff
@@ -51,28 +51,23 @@
...
+        csshttp.onreadystatechange = function () {
+            if (csshttp.readyState == 4 && csshttp.status == 200) {
+                return csshttp.responseText;
+            }
+        }
+        csshttp.open("GET", theURL, false);
+        csshttp.send();
+    }
 
     var config1 = {
```
asbjornu, line 72:
```suggestion
    var config = {
```

karlfristedt76: You're right

PR #94, opened 2020-06-14
Added missing failure message properties on class: SwedbankPay.Sdk.Pa…

SteffenSH:
…yments.Transaction

### DIFF
--- a/src/SwedbankPay.Sdk/Payments/Transaction.cs
+++ b/src/SwedbankPay.Sdk/Payments/Transaction.cs
@@ -14,6 +14,10 @@ namespace SwedbankPay.Sdk.Payments
                            Amount vatAmount,
                            string description,
                            string payeeReference,
+                           string failedReason,
+                           string failedActivityName,
+                           string failedErrorCode,
+                           string failedErrorDescription,
                            bool isOperational,
                            OperationList operations,
                            string activity)
@@ -28,6 +32,10 @@ namespace SwedbankPay.Sdk.Payments
             VatAmount = vatAmount;
             Description = description;
             PayeeReference = payeeReference;
+            FailedReason = failedReason;
+            FailedActivityName = failedActivityName;
+            FailedErrorCode = failedErrorCode;
+            FailedErrorDescription = failedErrorDescription;
             IsOperational = isOperational;
             Operations = operations;
             Activity = activity;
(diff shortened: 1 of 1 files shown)


asbjornu:
Thanks for this. Although it's not well documented, we should return a `problem` property containing an `application/problem+json` structure within it as an alternative to these legacy fields. We would rather expose `problem` than these `failedXyz` fields, as they will become less relevant over time as `problem` is the property we will focus our energy on.

Are you not seeing `problem` returned for failed transactions?

SteffenSH:
Hei Asbjørn!
Maybe I’m looking in the wrong places here, but I can’t find anywhere in the C# SDK/API to access this `problem` structure.
Yes, I can see the `problem` structure and these `failed...` fields in the JSON we get back…
E.g....

    {
      "transactionList": [
        {
          "failedReason": "ExternalResponseError",
          "failedActivityName": "Authorize",
          "failedErrorCode": "REJECTED_BY_ACQUIRER_CARD_STOLEN",
          "failedErrorDescription": "card stolen, response-code: 43",
          "isOperational": false,
          "problem": {
            "type": "https://api.payex.com/psp/errordetail/creditcard/acquirercardstolen",
            "title": "Operation failed",
            "status": 403,
            "detail": "Unable to complete Authorization transaction, look at problem node!",
            "problems": [
              {
                "name": "ExternalResponse",
                "description": "REJECTED_BY_ACQUIRER_CARD_STOLEN-card stolen, response-code: 43"
              }
            ]
          }
        }
      ]
    }

But how do I access them from the C# code?

If you plan to deprecate the `failed...` fields then I agree with you that it’s probably better to not implement those. But the `problem`  structure should at lest be exposed ?

asbjornu:
I'm glad you receive the `problems` property in the response, because that's the one we should both document on [developer.swedbankpay.com](https://developer.swedbankpay.com) (it's in our backlog and we'll get to it soon) as well as implement in this SDK. I would like to avoid both documenting and implementing the `failedXyz` properties, as they are legacy we want to fade away into the nether. 😃

If you can change this PR to remove the `failedXyz` properties and instead implement the `problems` property, I will happily accept and merge it. 🙏

SteffenSH:
I made an issue here: https://github.com/SwedbankPay/swedbank-pay-sdk-dotnet/issues/98

PR #240, opened 2021-01-29
Release v1.3.0

NullableInt:


asbjornu (changes requested):
We should do a beta on `develop` first and then this needs to bump to version `2.0` since it's a (massive) breaking change.

PR #243, opened 2021-02-01
Add meta data to infrastructure csproj

NullableInt:
This should fix the package name when we publish the `infrastructure` project.

### DIFF
--- a/src/SwedbankPay.Sdk.Infrastructure/SwedbankPay.Sdk.Infrastructure.csproj
+++ b/src/SwedbankPay.Sdk.Infrastructure/SwedbankPay.Sdk.Infrastructure.csproj
@@ -11,6 +11,8 @@
     <RepositoryType>git</RepositoryType>
     <PackageReleaseNotes>Swish payment handling. CreditCard payment handling. Checkout handling.</PackageReleaseNotes>
     <RootNamespace>SwedbankPay.Sdk</RootNamespace>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/SwedbankPay.Sdk.Infrastructure/SwedbankPay.Sdk.Infrastructure.csproj
+++ b/src/SwedbankPay.Sdk.Infrastructure/SwedbankPay.Sdk.Infrastructure.csproj
@@ -13,7 +13,6 @@
     <RootNamespace>SwedbankPay.Sdk</RootNamespace>
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);NugetPack</TargetsForTfmSpecificBuildOutput>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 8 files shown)


asbjornu (approved):

src/.nuspec
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <!-- Required elements-->
+        <id>SwedbankPay.Sdk</id>
+        <description>A client library and SDK for Swedbank Pay's RESTful API Platform</description>
+        <authors>Swedbank Pay</authors>
+        <version>1.1.1</version>
```
asbjornu, line 8: Is it possible to not hard-code `version` in the `.nuspec` and instead just provide it on the command line as minted by GitVersion?

NullableInt: Yes, already done

asbjornu: But the `version` element still needs to be in the `.nuspec` file?

NullableInt: Yes, it needs to be there, but it gets overwritten from the CLI. I've tested it locally to ensure it gets overwritten.

asbjornu: Is it possible to set the version number to something invalid so version `1.1.1` is not accidentally published anywhere?

```suggestion
        <version>0.0.1-INVALID</version>
```

src/SwedbankPay.Sdk.Infrastructure/SwedbankPay.Sdk.Infrastructure.csproj
```diff
@@ -3,6 +3,13 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.1</TargetFramework>
     <LangVersion>8.0</LangVersion>
```
asbjornu, line 5: Shouldn't `LangVersion` be the same in `Core` and `Infrastructure`?

src/.nuspec
```diff
@@ -0,0 +1,21 @@
...
+        <!-- Optional elements -->
+        <projectUrl>https://github.com/SwedbankPay/swedbank-pay-sdk-dotnet</projectUrl>
+        <license type="expression">Apache-2.0</license>
+        <icon>icon.png</icon>
+        <copyright>Swedbank Pay</copyright>
+        <tags>SwedbankPay sdk payments payment client api rest restful</tags>
+        <language>en-US</language>
+    </metadata>
+    <files>
+        <file src="../icon.png" target="" />
```
asbjornu, line 19: The `icon` element above references `./icon.png`, while this `file` element references `../icon.png`. Which one is correct? And is it necessary to explicitly include `icon.png` when it is already referenced with `<icon/>`?

src/.nuspec
```diff
@@ -0,0 +1,21 @@
...
+
+        <!-- Optional elements -->
+        <projectUrl>https://github.com/SwedbankPay/swedbank-pay-sdk-dotnet</projectUrl>
+        <license type="expression">Apache-2.0</license>
+        <icon>icon.png</icon>
+        <copyright>Swedbank Pay</copyright>
+        <tags>SwedbankPay sdk payments payment client api rest restful</tags>
+        <language>en-US</language>
+    </metadata>
+    <files>
```
asbjornu, line 18: Shouldn't any of the project files or DDLs be referenced by the `.nuspec` file?

asbjornu (changes requested):

src/.nuspec
NullableInt: Yes, it's necessary. https://docs.microsoft.com/en-us/nuget/reference/nuspec#icon

src/.nuspec
NullableInt: I did not find any such parameter to pass inn. And the project file has to be provided when running `dotnet restore`

asbjornu: As long as the published NuGet works, it's fine. Can we add a `nuget publish` and `nuget install` step in GitHub Actions so we know it all works as it should?

NullableInt: I'll get on that

asbjornu (approved):

PR #246, opened 2021-02-05
Update what we pack

NullableInt:
This avoids us packing `SwedbankPay.Sdk.dll` twice


asbjornu (approved):
I don't really understand how this is going to work, but okay.

PR #248, opened 2021-02-08
Feature/set target framework to netstandard 20

NullableInt:


src/SwedbankPay.Sdk.Infrastructure/Extensions/EnumExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SwedbankPay.Sdk.Extensions
+{
+    public static class EnumExtensions
+    {
+        public static TEnum Parse<TEnum>(String value) where TEnum : struct
```
asbjornu, line 9: Make this into a proper extension method so the word `Extensions` make sense.

```suggestion
    public static class StringExtensions
    {
        public static TEnum ToEnum<TEnum>(this string value) where TEnum : struct
```

asbjornu (changes requested):
If we're building for several different framework versions, we need to introduce `matrix` into the GitHub Actions workflow, no? Also, the tests are failing.

NullableInt:
> 
> 
> If we're building for several different framework versions, we need to introduce `matrix` into the GitHub Actions workflow, no? Also, the tests are failing.

Where do you want matrix testing? And on what?

Test failed due to `{"name":"Transactions[1].Activities","description":"Activity chain is broken."}` meaning we have to re run the test after a few minutes. This is not an error in production and only happens on external integration.

src/SwedbankPay.Sdk.Infrastructure/PaymentInstruments/Card/CardPayment.cs
```diff
@@ -22,7 +22,7 @@ public CardPayment(CardPaymentDto payment)
             Currency = new Currency(payment.Currency);
             Description = payment.Description;
             Id = new Uri(payment.Id, UriKind.RelativeOrAbsolute);
-            Instrument = EnumExtensions.Parse<PaymentInstrument>(payment.Instrument);
+            Instrument = StringExtensions.Parse<PaymentInstrument>(payment.Instrument);
```
asbjornu, line 25:
```suggestion
            Instrument = payment.Instrument.ParseTo<PaymentInstrument>();

```

src/SwedbankPay.Sdk.Infrastructure/Extensions/StringExtensions.cs
```diff
@@ -4,9 +4,10 @@
 
 namespace SwedbankPay.Sdk.Extensions
 {
-    public static class EnumExtensions
+    public static class StringExtensions
+
     {
-        public static TEnum Parse<TEnum>(String value) where TEnum : struct
+        public static TEnum Parse<TEnum>(this string value) where TEnum : struct
```
asbjornu, line 10:
```suggestion
        public static TEnum ParseTo<TEnum>(this string value) where TEnum : struct
```

src/SwedbankPay.Sdk.Infrastructure/PaymentInstruments/Card/CardPayment.cs
```diff
@@ -43,7 +43,7 @@ public CardPayment(CardPaymentDto payment)
             Transactions = payment.Transactions?.Map();
             if (!string.IsNullOrEmpty(payment.Intent))
             {
-                Intent = EnumExtensions.Parse<PaymentIntent>(payment.Intent);
+                Intent = StringExtensions.Parse<PaymentIntent>(payment.Intent);
```
asbjornu, line 46:
```suggestion
                Intent = payment.Intent.ParseTo<PaymentIntent>();

```

asbjornu (changes requested):
>Where do you want matrix testing? And on what?

[`matrix`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) is used to parameterize every `job` for a different global variable, usually specifying the environment.

We should perform `dotnet build`, `dotnet test`, etc., for each version of .NET we claim to support, no?

NullableInt:
> 
> 
> > Where do you want matrix testing? And on what?
> 
> [`matrix`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) is used to parameterize every `job` for a different global variable, usually specifying the environment.
> 
> We should perform `dotnet build`, `dotnet test`, etc., for each version of .NET we claim to support, no?

I can set this up for the test step. But each run will take approximately 19 minutes.

NullableInt:
> 
> 
> > > Where do you want matrix testing? And on what?
> > 
> > 
> > [`matrix`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) is used to parameterize every `job` for a different global variable, usually specifying the environment.
> > We should perform `dotnet build`, `dotnet test`, etc., for each version of .NET we claim to support, no?
> 
> I can set this up for the test step. But each run will take approximately 19 minutes for each configuration.

Looking into it this will not work as the test projects does not use the nuget (nor should they), so I'll need to come up with a different way of testing this.

NullableInt:
@asbjornu Can we do matrix testing in a different pull request?

asbjornu:
> @asbjornu Can we do matrix testing in a different pull request?

Yes, absolutely.

asbjornu (approved):

PR #277, opened 2021-02-24
Card verificationlist

NullableInt:


src/SwedbankPay.Sdk/PaymentInstruments/Card/IVerifyTransaction.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Textual reason for failure code if there is one.
+        /// </summary>
+        string FailedErrorCode { get; }
+
+        /// <summary>
+        /// Textual description for failure if there is one.
+        /// </summary>
+        string FailedErrorDescription { get; }
```
asbjornu, line 28: We should use `Problem` instead of the `Failed…` properties.

```suggestion
        IProblem Problem { get; }
```

asbjornu (changes requested):

src/SwedbankPay.Sdk/PaymentInstruments/Card/IVerifyTransaction.cs
NullableInt: I don't controll the API response

asbjornu (changes requested):
One test for `problem` within a `transaction` and one for `problem` within a `transaction` within a `transactionList` would be good. Otherwise, this looks great!

NullableInt:
> 
> 
> One test for `problem` within a `transaction` and one for `problem` within a `transaction` within a `transactionList` would be good. Otherwise, this looks great!

Is the test I added good?

asbjornu (approved):

PR #278, opened 2021-03-02
Feature/update build test publish

NullableInt:
A more complete version of #274 
Fixes so the workflow runs on push and pull requests. As well as making the casing consistent (and hopefully correct) for NUGETVERSIONV2.


asbjornu:
I don't think this is required anymore.

PR #303, opened 2021-03-23
Feature/fix empty salelist response

zunkas:
Fix for saleList not being populated

### DIFF
--- a/src/SwedbankPay.Sdk.Infrastructure/PaymentInstruments/Swish/SaleListItemDto.cs
+++ b/src/SwedbankPay.Sdk.Infrastructure/PaymentInstruments/Swish/SaleListItemDto.cs
@@ -2,14 +2,25 @@
 
 namespace SwedbankPay.Sdk.PaymentInstruments.Swish
 {
-    internal class SaleListItemDto
-    {
-        public DateTime Date { get; }
-        public Uri Id { get; }
-        public string PayerAlias { get; }
-        public string PaymentRequestToken { get; }
-        public string SwishPaymentReference { get; }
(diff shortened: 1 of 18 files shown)


src/SwedbankPay.Sdk.Infrastructure/PaymentInstruments/Swish/SaleListItemDto.cs
```diff
@@ -2,14 +2,25 @@
...
+			Transaction = transaction;
+		}
+
+		public DateTime Date { get; }
+		public Uri Id { get; }
+		public string PayerAlias { get; }
+		public string PaymentRequestToken { get; }
+		public string SwishPaymentReference { get; }
+		public string SwishStatus { get; }
+		public TransactionDto Transaction { get; }
```
asbjornu, line 24: Since I assume this DTO is used for deserialization, is it not better to just make the properties have public getters instead of adding a constructor? DTO objects are only for data transfer, after all.

asbjornu (changes requested):

src/SwedbankPay.Sdk.Infrastructure/PaymentInstruments/Swish/SaleListItemDto.cs
zunkas: public setters you mean?

asbjornu: Yes, sorry.

asbjornu (approved):

PR #392, opened 2021-08-25
Moved RiskIndicator and made it public

zunkas:


src/SwedbankPay.Sdk/RiskIndicator.cs
```diff
@@ -1,66 +1,65 @@
...
-        ///     02 (Ship to another verified address on file with merchant)
-        ///     03 (Ship to address that is different than cardholder's billing address)
-        ///     04 (Ship to Store / Pick-up at local store.Store address shall be populated in shipping address fields)
-        ///     05 (Digital goods, includes online services, electronic giftcards and redemption codes)
-        ///     06 (Travel and Event tickets, not shipped)
-        ///     07 (Other, e.g.gaming, digital service)
-        /// </summary>
-        public ShipIndicator ShipIndicator { get; set; }
-    }
+﻿using SwedbankPay.Sdk.PaymentOrders;
```
asbjornu, line 1: This `using` statement feels weird. If `RiskIndicator` requires stuff from `PaymentOrders`, we either need to keep `RiskIndicator` inside the `PaymentOrders` namespace or we have to move whatever it depends on out of `PaymentOrders`.

asbjornu (changes requested):

src/SwedbankPay.Sdk/RiskIndicator.cs
zunkas: Moved the dependent classes from `PaymentOrders` namespace. Since `RiskIndicator` is being used with PaymentInstruments also this seems more logical

PR #457, opened 2022-03-14
Added default request header for user-agent

zunkas:


perhassle (approved):

src/SwedbankPay.Sdk/UserAgent.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Reflection;
+
+namespace SwedbankPay.Sdk
+{
+    public static class UserAgent
+    {
+        public static string Default;
+        static UserAgent()
+        {
+            Default = $"swedbankpay-sdk-dotnet/{Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion}";
```
asbjornu, line 10: I suspect `InformationalVersion` is going to be a bit too verbose. If GitVersion is responsible for providing its value, it's going to contain the Git SHA, for instance. I think `AssemblyFileVersionAttribute` is sufficient. I also think we should have a test that verifies that the value returned here contains a sensible version number and not just the default `1.0.0.0`.

asbjornu (changes requested):
I like that `userAgent` is being made optional, but can we somehow make it so the `SwedbankPayClient` constructor is the only place where `userAgent` is needed and setting it there propagates it to every request? Repeating it in every `Details` request seems a bit superfluous to me.

PR #549, opened 2022-11-15
Implement V3 redirect checkout

nicoletacristian:


src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderRequest.cs
```diff
@@ -27,10 +28,11 @@ public class PaymentOrderRequest
                                    Language language,
                                    bool generateRecurrenceToken,
                                    IUrls urls,
-                                   IPayeeInfo payeeInfo)
+                                   IPayeeInfo payeeInfo,
+                                   bool useCheckoutV3 = false)
```
asbjornu, line 32: I feel that `bool useCheckoutV3` does not scale very well. What should we do when we launch Checkout v4? I would like to explore other designs for this, perhaps by adding a `PaymentOrderRequestBuilder` which can reduce the number of arguments in `PaymentOrderRequest` as well as make it immutable.

asbjornu (changes requested):
Thank you so much for this PR! It's a great start, but we should probably think a bit more about how we want Checkout v3 implemented in the SDK. It may be worth doing some backwards incompatible changes for. Thoughts @zunkas?

src/SwedbankPay.Sdk/PaymentOrders/PaymentOrderRequest.cs
mathiso: Can you please just resolve this and complete this PR? Swedbank does not want us to use Checkout v2 any more.

zunkas:
A new version of the SDK for version 3.1 have been merged to develop.

PR #572, opened 2023-03-02
Update MobilePayPaymentAuthorizationDto.cs

antonsehovac:
Properties aren't getting mapped since they doesn't have any setters


antonsehovac:
@asbjornu Could you please review this bug?

zunkas (approved):

Issue #619, opened 2024-03-12
SwedbankPay.Sdk.Extensions -> SwedbankPay.Sdk.NotPacked?


JeppePepp:
Hello,

There is a dependency on the SwedbankPay.Sdk.Extensions package for SwedbankPay.Sdk.NotPacked >= 4.0.0  that Im not able to resolve.

Error from a new project when trying to install after installed SwedbankPay.Sdk 4.0.0

Severity	Code	Description	Project	File	Line	Suppression State
Error	NU1101	Unable to find package SwedbankPay.Sdk.NotPacked. No packages exist with this id in source(s): C:\Program Files\dotnet\library-packs, Microsoft Visual Studio Offline Packages, nuget.org	TestSdk	C:\GIT\Dev\TestSdk\TestSdk.csproj	1

PR #628, opened 2024-03-19
Update project reference properties in SwedbankPay.Sdk.Extensions.csproj

zunkas:
This commit modifies the SwedbankPay.Sdk.Extensions.csproj to change the properties of the project reference to SwedbankPay.Sdk. Now, all the project's assets are marked as private to ensure they are not exposed or shared with other projects unintentionally.


lounge:
#619

lounge (approved):

